PR #4, opened 2018-12-17
WIP: add basic SVG rendering support

flosse:
It would be cool to be able to render SVG elements as well.
Here is a first draft as a base for a discusstion :)


David-OConnor:
SVG functionality would be great. I don't know much about it, but it looks like you've got a solid start. I'm working on adding fetch and routing now.

PR #7, opened 2018-12-18
add all possible SVG tags

flosse:


flosse:
Attention: The macros for the SVG elements are still missing, because I didn't want to copy/paste the same code again and again. And I was not able yet to create a general macro for that :-\

flosse:
> For now, we may need lots of repetitive code. Sounds like it's biting you here too on SVG elements.

ok, here are some more svg macros: #9

PR #18, opened 2019-01-04
Completely refactoring fetching

Pauan:


limira:
Very great code. I wish that I had a skill level like you. I am never happy with my code. Do you mind if I apply similar things them to my project?

But I think examples need to be updated as well. For example, `examples/server_interaction` use both `get_json` and `post_json` which a now gone in your PR!

@David-OConnor, I think that Seed need more tests. At least, required contributors ensure that every examples successfully built before requesting a merge? What do you think?

src/fetch.rs
```diff
@@ -48,191 +47,144 @@ impl Method {
...
-            // https://rustwasm.github.io/wasm-bindgen/api/web_sys/enum.RequestMode.html
-            mode: web_sys::RequestMode::Cors,
+            url,
+            init: web_sys::RequestInit::new(),
+            headers: web_sys::Headers::new().expect("Error with creating Headers"),
         }
     }
-}
 
+    #[inline]
```
limira: This is more a question than a review. Do we really need all these #[inline]? Will it eventually be optimized away if we `inline` them?

Pauan: I'm not an expert on low-level optimizations, but my understanding is: in LTO mode Rust + LLVM will do a lot of inlining automatically, so the `#[inline]` is ***not*** necessary.

But in non-LTO mode the `#[inline]` can help with performance.

So I just got into the habit of always adding `#[inline]` to small functions which should be inlined, to help guarantee that they are inlined.

limira: So, I think in a wasm framework like this, performance should be focus on the DOM render/update. In other part of the framework, I think that we should focus optimizing on size instead of performance?

Pauan: Inlining (when done properly) helps with both performance and code size, since all of the function wrappers and abstractions melt away, and it enables better optimizations by the compiler.

This is Rust, we don't need to make compromises: we can make *everything* fast, including the DOM, but also including everything else too.

Pauan:
> Very great code. I wish that I had a skill level like you.

Thanks! I appreciate that. A lot of it is just practice and knowledge.

> Do you mind if I apply similar things them to my project?

Of course, go ahead. They're good techniques which should be commonly used.

Nobody has (or should have) a monopoly on ideas.

> But I think examples need to be updated as well.

Oh, you're right! Let me fix that. Thanks for pointing it out.

src/fetch.rs
```diff
@@ -48,191 +47,144 @@ impl Method {
...
-        .and_then(move |v| {
-            callback(v);
-            future::ok(JsValue::null())
-        });
-
-    wasm_bindgen_futures::future_to_promise(f);
+    pub fn fetch_json<A: DeserializeOwned>(self) -> impl Future<Item = A, Error = JsValue> {
+        self.fetch_string()
+            .map(|text| {
+                serde_json::from_str(&text).expect("Error deserializing JSON")
```
jesskfullwood: Not sure what the error-handling story should be with this API, but this `expect` in particular would seem to panic if the server returns malformed data, which is not acceptable. I think instead the error should be propagated inside the `Future`

naufraghi:
@Pauan can you please summarize what the refactoring is about and what are the pieces that are to be reviewed with more attention?

PS I'd like to help on this project (thanks @David-OConnor and feel free to override my request), this infos is something I usually ask for each PR to easy the review process.

David-OConnor:
@naufraghi For context, I've been attempting to get a fetch API working for a few weeks, and posted an issue in the wasm-bindgen repo requesting more detailed docs/examples. The fetch implementation I committed yesterday, which @Pauan is replacing, had inelegant code, and a non-ideal API; it was the only solution I could get working, rather than the one I wanted. This PR addresses this through better use of Futures.

David-OConnor:
@limira Agree. I've held off on tests so far due to how fast the API's moving; they add friction at early stages due to having to edit tests frequently. We're at the point now where they make sense.

src/fetch.rs
Pauan: I agree in principle (I'm hugely ***not*** a fan of `unwrap`/`expect`), however `JsFuture` has an error type of `JsValue`, whereas Serde does not. So it'll require some thinking on how to convert between them.

Pauan:
@David-OConnor Please don't merge this yet. I still need to get the examples fixed.

Pauan:
@David-OConnor I updated the examples. I think this is ready for merging now.

PR #25, opened 2019-01-21
Return App from seed::run

flosse:
Modifying the app state from the outside
could be helpfull, eg.:

```rust
pub fn render() {
    let app = seed::run(Model::default(), update, view, "app", None, None);
    open_websockets(app);
}

fn open_websockets(state: seed::App<Msg, Model>) {

  // setup websockets ...

  let on_message = Box::new(move|ev: MessageEvent| {
    let txt = ev.data().as_string().unwrap();
    let json = : JsonMsg = serde_json::from_str(&text).unwrap();
    state.update(Msg::Json(json));
  });
}
```


David-OConnor:
Nice, and nonbreaking

Issue #8, opened 2018-12-18
WebSocket support & example


flosse:
It would be cool to create an example that uses WebSocket :)

David-OConnor:
Agree. It would be appropriate for the WIP server_integration example.  First, I'm going to first attempt basic fetch requests: this is a critical missing feature. Asking an API user to dive into the raw web_sys fetch/Request API's not feasible. Then would love WebSocket support.

flosse:
related: #38

thedodd:
I've recently built an app with this framework (which is awesome btw) and I used the websockets example to get started with setting up the Websockets integration. I wanted to share a few thoughts on my experience to see what you all think in terms of potential improvements &c.

- Using wasm-bindgen types to differentiate between string & binary messages was a bit of a hurdle. I got it working, but I definitely see this as a stumbling block for new users. Could be fairly defeating for some folks I suspect.
- I'm thinking we could introduce a new Websocket type which wraps a standard wasm-bindgen / web-sys Websocket.
- This new type could introduce a few convenience methods for being able to handle binary vs text messages.
- Additionally, we could use this higher-level type to store the underlying web-sys Websocket as well as its callbacks. Finding a way to store the callbacks without having to leak all of them (which is obvs not good) could be a stumbling block for folks.
- We could simply store all of the callbacks in a private vec on the wrapper Websocket type. When it goes out of scope, the callbacks will be dropped as well.
- The callbacks can be simply stored in an `Rc` wrapped inside of an enum to account for function type variance. This is what I did. Worked out pretty well.

Anyway, @David-OConnor you're an absolute champion. Thanks for all of the excellent work you've put into this. This framework is a game-changer, no doubt. I'm happy to help out in any way.

David-OConnor:
Agree - I think the way forward is to add (and document) a high-level WS API. The example's a low-level stopgap that demonstrates it's possible, and shows the `web_sys`-level API endpoints we could use when building this API. Complicating factor: I haven't used WS, so am not (yet?) in a position to comment on the details. Want to submit a PR of your path forward? It doesn't have to be complete, documented or finalized: Any API that wraps the web_sys code, and addresses problems you ran into is better than the current approach (ie none).

thedodd:
[...]

PR #28, opened 2019-01-22
WIP: add websocket example

flosse:
This is a first try to solve #8.
It's incomplete at the moment. I'm very happy to receive some feedback :)


David-OConnor:
I've never used WS, but will take a look this week. I suspect we can incorporate much of this into the framework itself, providing a high-level API.

David-OConnor:
Merging this now; will investigate incorporation later

PR #29, opened 2019-01-23
Some ideas

flosse:
I dislike using a `run` function with so many arguments.
Here is an approach that might be a bit cleaner.
What do you think?


David-OConnor:
I like this! When I initially set up the run interface, I intended it for use with only 4 params, which didn't seem excessive. It's grown to 6 with routing and window events, which has crossed the line. The .routes and .window_events methods make sense as a way to add optional params, but what adv do you see to separating .mount?

flosse:
> but what adv do you see to separating .mount?

It's maybe an edge case but you could run multiple instances of e.g. an "widget" at different places on a page.

https://github.com/David-OConnor/seed/blob/671f0750ee76b6c4c0b04e15bd740ca876540871/examples/counter/src/lib.rs#L127-L130

But of course we can ignore this case.

flosse:
...and we can just define a default mount mount:

https://github.com/David-OConnor/seed/blob/28cb18a4b8969f33e4f0b6c57fb03444e2ca5671/src/vdom.rs#L92

flosse:
You don't have to merge this PR ... it's more like a base for discussing changes. If you like one or more of them I can create specific PRs to get a cleaner git history :)

David-OConnor:
That solves my verbosity concern, and I like the rest of the changes

David-OConnor:
I'll update the docs and make a new release with these changes once I take another hack at dynamic routing

PR #38, opened 2019-01-29
Ws with listener

flosse:
This is a quick and dirty approach to be able to send messages from the client to the server.
Do you have an idea how to do it better?


David-OConnor:
Given I've never used websockets... nope! Still would like to abstract more of this into the framework directly, and document once done.

PR #48, opened 2019-02-12
A new solution to the raw_html problem, append to parents inner_html

tiberiusferreira:
This now works with:

```
El::from_html("<div>
                <p>Example: This is a <strong>sentence</strong> of text and it also has some <em>italicized text</em> too.</p>
                <p>This is another paragraph.</p>
                </div>")
```
Resulting in:
screen shot 2019-02-11 at 22 04 40

However, it works by attaching the ```<span>``` raw_html to the parent nodes inner_html. The ```<span>``` children (is there a way to create them?) are also attached directly to the parent. 

Is this what you had in mind?

### DIFF
--- a/src/websys_bridge.rs
+++ b/src/websys_bridge.rs
@@ -3,7 +3,7 @@ use wasm_bindgen::JsCast;
 
 use crate::dom_types;
 use crate::dom_types::El;
-
+use web_sys::HtmlElement;
 /// Reduces DRY
 /// todo can't find a suitable trait for this. Seems like set_autofocus is
 /// implemented individually for each of these el types.
@@ -134,9 +134,6 @@ pub fn make_websys_el<Ms: Clone>(
                 .expect("Missing text on raw HTML element"),
         );
 
-        // We know that there is a child because we just attached it
-        let el_ws = el_ws.first_element_child().unwrap();
-
         // todo DRY
         if el_vdom.style.vals.keys().len() > 0 {
             el_ws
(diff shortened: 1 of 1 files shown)


David-OConnor:
Looks good. I attempted a different approach, with this code inserted in `attach_el_and_children`:
```rust
   if el_vdom.raw_html {
        let html_children = el_ws.child_nodes();
        crate::log(html_children.length());

        for i in 0..html_children.length() {
            let child_in_span = html_children.item(i)
                .expect("Missing child in raw html element");

            el_ws.remove_child(&child_in_span)
                .expect("Problem removing child from span in raw html element");
            parent.append_child(&child_in_span)
                .expect("Problem appending child in raw html element");
        }
    }
        else {
            parent.append_child(&el_ws).expect("Problem appending child");
        }
```
Was unable to get it working due to nondescript runtime errors.

tiberiusferreira:
The problem with the code you posted might be the following:

Imagine you have 2 children. So the for goes item(0) and item(1). 
item (0) ; item (1);
When you remove the item (0), item (1) becomes item(0):
item(0)
Then you try to remove item(1), but it does not exist. 

I have your solution (which is way cleaner) working on my fork. I'll make a PR.

tiberiusferreira:
Oh, it looks like you are already have some work in progress going on in there, so I won't make a PR. 

Anyway thanks for your work in this project, really appreciate it.

David-OConnor:
Hey - I merged in yours. Works for initial render, but appears not to update on changes. I cleaned up the merge conflicts - feel free to do a PR; I'll hold off on changes for now.

PR #49, opened 2019-02-12
revised solution for span raw html problem

tiberiusferreira:
This is my attempt at your proposed solution. 
About your comment on the previous attempt
 "Works for initial render, but appears not to update on changes." 
you mean it does not render correctly on router changes for example?

### DIFF
--- a/src/websys_bridge.rs
+++ b/src/websys_bridge.rs
@@ -203,32 +203,6 @@ pub fn attach_el_and_children<Ms: Clone>(el_vdom: &mut El<Ms>, parent: &web_sys:
 
     let el_ws = el_vdom.el_ws.take().expect("Missing websys el");
 
-//<<<<<<< HEAD
-//    // Don't attach if raw_html; these are initially wrapped in a span tag. We'll
-//    // extract the children, and attach those instead in the looop below.
-//    if el_vdom.raw_html {
-//        let html_children = el_ws.child_nodes();
-//        crate::log(html_children.length());
-//
-//        for i in 0..html_children.length() {
-//            let child_in_span = html_children.item(i)
-////            let child_in_span = html_children.get(i)
-//                .expect("Missing child in raw html element");
-//
-//            el_ws.remove_child(&child_in_span)
-//                .expect("Problem removing child from span in raw html element");
-//            parent.append_child(&child_in_span)
-//                .expect("Problem appending child in raw html element");
-//        }
-//    }
-//        else {
-//            parent.append_child(&el_ws).expect("Problem appending child");
-//        }
-//
-//    for child in &mut el_vdom.children {
-//        attach_el_and_children(child, &el_ws)
-//=======
-
 
     if !el_vdom.raw_html {
         // appending the element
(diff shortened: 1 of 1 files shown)


David-OConnor:
> you mean it does not render correctly on router changes for example?
Correct. I haven't looked into the details or done any troubleshooting, but a test on the homepage shows the guide page text not updating.

tiberiusferreira:
> Great find on fixing the prob with my code.

 Thanks! 
> edit: A glance at this PR shows the same issue; I think we need to update some patching code too.

Yes, I cloned the homepage example repo and can reproduce it. If I find a fix I'll let you know.

tiberiusferreira:
The weird part is that routing works fine when changing between Guide and Changelog, but not between Guide pages.

David-OConnor:
Yeah - likely because we're switching between an HTML el and a non-HTML el, which have diff tags, and therefore force a full rerender.

I'm examining a different approach, where the `from_html` and `from_markdown` funcs create full vdom Els, instead of handling it during render.

tiberiusferreira:
Well, I think I found the problem:
https://github.com/David-OConnor/seed/blob/85aa90d22fda179e00c2bd1eab9b2dbb300b9404/src/websys_bridge.rs#L332-L337 
Here we set the inner_html of the element actual DOM node, which is the ```<span>``` we did not actually attach to the DOM during first rendering. 

I tried to fix it by keeping references of the nodes which we attached during the first render and modifying those, but it got complicated and messy. The real problem is the vdom and actual DOM not in sync anymore.

I think the best solution is what you are examining, creating full vdom Els from `from_html` and `from_markdown`. 

I imagine it would involve parsing the html string passed. If so, long term this could be done at compile time and provide an additional check of correctness and help catch typos and invalid html.

David-OConnor:
Nice find again. Don't need to manually parse; can set inner_html on a dummy el, then create a nested vdom node from it. Almost have it working.

tiberiusferreira:
Thanks for the changes.

 After trying it to render some SVGs, it had some problems. (It would not render anything).
I thought it was because iterating over a hashmap to set the attributes does not preserve ordering and it might affect it. So I tried changing the hashmap to https://github.com/bluss/indexmap .

Now the attributes order is preserved, but it still does not render (while before it rendered fine). 

I don't know yet the source of the problem, but if I find it, I will let you know. 

Here is the code I was testing with: https://github.com/tiberiusferreira/festera_seed

David-OConnor:
Still haven't figured out what's wrong with the latest version. For the homepage, it alternates between several errors when switching pages.

David-OConnor:
The latest commit works for the homepage, although I still don't understand the root cause of the problems: Instead of forcing a panic, it now recovers, and outputs a note in the console. Is this working for you?

Is the SVG problem you mentioned a separate issue?

tiberiusferreira:
Unfortunately it's not working for me :( 
Using Commit 85aa90d22fda179e00c2bd1eab9b2dbb300b9404 I get:
screen shot 2019-02-16 at 23 36 58

and using latest ( bd90f1cb15c9fca98398a63019d9829ce6d1a295 ) I get: 

screen shot 2019-02-16 at 23 36 08

Once I get some time I will try to help pinpoint the issue. The code I'm testing against is this one https://github.com/tiberiusferreira/festera_seed

David-OConnor:
Weird; that was intermediate gibberish text I added to debug that shouldn't be in the commit.

David-OConnor:
Two observations: It appears to be ignoring the `<g>` tag. I'm not sure if the xml metadata is being ignored as well.

tiberiusferreira:
Just tried with 
```
[dependencies]
seed = {git = "https://github.com/David-OConnor/seed.git", rev = "bd90f1cb15c9fca98398a63019d9829ce6d1a295"}
```
and now the MOO text is gone, but the SVG is still missing:

screen shot 2019-02-17 at 00 14 49

Thanks for looking into it.

David-OConnor:
More info: The SVG bug happens with or without Raw HTML, and if you edit the SVG tag in the Firefox console.... any edit at all, and it will render.

David-OConnor:
Fixed in latest commit. The problem was not setting the namespace. There still appears to be a problem with creating SVGs using macros.

tiberiusferreira:
Yep, it works! Thank you very much.

PR #57, opened 2019-02-28
Avoid cloning model in update()

sapir:


David-OConnor:
Great stuff - I've been looking for a way to avoid requiring `Clone` on the model for a while, but had given up.

PR #62, opened 2019-03-04
More elements

sapir:


David-OConnor:
Great cleanup of the shortcut module. I'd assumed this would have to wait for a future version of Rust, or proc macros

PR #63, opened 2019-03-04
Include URL search and hash in push_state_with_url() call

sapir:
I tried to implement "hash routing" but the hash part of the route wasn't being included in some cases. I think this helped at least partially, but I'm not sure.


David-OConnor:
Need to do some research on what  location.set_hash and set_search actually do, since I haven't used either in a proj. I suspect this approach is fine.

PR #64, opened 2019-03-05
Fix empty insertion position

### LINKED ISSUE #60: Empty changed to El gets inserted at incorrect position
I'll make a pull request shortly with a test showing what I mean

sapir:
Fixes #60

### DIFF
--- a/src/vdom.rs
+++ b/src/vdom.rs
@@ -533,23 +533,21 @@ fn patch<Ms: Clone>(
                 // todo: Perhaps some of this next segment should be moved to websys_bridge
                 websys_bridge::attach_children(new);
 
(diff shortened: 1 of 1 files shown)


David-OConnor:
Nice. Looks like the problem was naively appending a child in this code:
```rust
                if old.empty {
                    // When adding an empty element, there's no existing web_sys Element to
                    // replace, so we add it in the proper sequence of its parents' children.
                    parent.append_child(&new_el_ws)
                        .expect("Problem adding element to previously empty one");
                } else {
```
I had a WIP attempt where I'd try to find the proper insertion point in the children loop, but your approach is nicer

PR #66, opened 2019-03-05
Remove El.id and .nest_level

sapir:


sapir:
I think that because of how the diffing works right now, they don't actually add any value?

David-OConnor:
Agree. I suspect id might make an appearance later, but for now, dead code.

PR #67, opened 2019-03-06
Delay websys el creation

sapir:
I *think* commit 15fffb8 here improves performance noticeably for updates with lots of elements where most of the elements don't change at all, but my testing isn't too scientific, so I'm not entirely sure. 

btw, would you mind running cargo fmt?


David-OConnor:
I especially like the `walk_tree_mut` function. Haven't checked performance yet.

PR #69, opened 2019-03-06
&mut for Update + returning Futures from Update

sapir:
My rationale for the suggested changes follows.

### Rationale for passing the model parameter by `&mut`

In my app I have a few subcomponents, so the update function for my parent component looks something like this:
```rust
match message {
  Msg::InnerComponent1(message) => {
    let Model { inner1, inner2, inner3 } = model;
    let upd = inner_module::update(message, inner1);
    wrap_update(
      upd,
      |inner1| Model { inner1, inner2, inner3 },
      |msg| Msg::InnerComponent1(msg),
    )
  }
}
```
where `wrap_update` is a helper function that matches the contents of the Update enum and converts the model part to the parent component's model using the first closure, and also converts the message part to the parent component's message type using the second closure.

Anyway, I have to unpack my Model and then put it back together so that I can pass the inner component's Model to its `update` function. I think using a mut reference might work better here. It's a bit different than the way Elm does it, but then Elm is a pure language, and doesn't have any mut references anywhere (or ownership rules, either).

With a mutable reference, it would look like this:
```rust
match message {
  Msg::InnerComponent1(message) => {
    let upd = inner_module::update(message, &mut model.inner1);
    wrap_update(upd, |msg| Msg::InnerComponent1(msg))
  }
}
```

### Rationale for making Update a struct

This is so it's easier to change the parts separately (like in my `Update::map` method, which is basically just the `wrap_update` function I mentioned earlier + support for futures). I did try to add some convenience traits/methods to make things easier, so although you can't just return `Render` any more, you can at least return `Render.into()`.

My previous example would now look like this:
```rust
match message {
  Msg::InnerComponent1(message) => {
    inner_module::update(message, &mut model.inner1)
      .map(|msg| Msg::InnerComponent1(msg))
  }
}
```

### Rationale for triggering futures in update

1. Sometimes I want to trigger a fetch request from `update` without having been called from `view` with that purpose.
2. Having to put the state in the message built in the view function gets more complicated when the message belongs to an inner component.
3. Having access to the app state in the `view` function seems a bit out of place, as does deciding on whether to put the app state inside the message or not - That would depend on whether an effect should happen, which is `update`'s job.
4. This is inspired by Elm's `Cmd` and `Task`.

### DIFF
--- a/examples/counter/src/lib.rs
+++ b/examples/counter/src/lib.rs
@@ -30,33 +30,23 @@ enum Msg {
     ChangeWWC(String),
 }
 
-/// The sole source of updating the model; returns a fresh one.
-fn update(msg: Msg, model: Model) -> Update<Msg, Model> {
+/// The sole source of updating the model
+fn update(msg: Msg, model: &mut Model) -> Update<Msg> {
     match msg {
-        Msg::Increment => Render(Model {
-            count: model.count + 1,
-            ..model
-        }),
-        Msg::Decrement => Render(Model {
-            count: model.count - 1,
-            ..model
-        }),
-        Msg::ChangeWWC(what_we_count) => Render(Model {
-            what_we_count,
-            ..model
-        })
+        Msg::Increment => {
+            model.count += 1;
+            Render
+        }
+        Msg::Decrement => {
+            model.count -= 1;
+            Render
+        }
+        Msg::ChangeWWC(what_we_count) => {
+            model.what_we_count = what_we_count;
+            Render
+        }
     }
 }
-
-/// A mutable-style alternative:
-//fn update(msg: Msg, model: mut Model) -> ModelUpdate<Model> {
-//    match msg {
-//        Msg::Increment => model.count += 1,
-//        Msg::Decrement => model.count -= 1,
-//        Msg::ChangeWWC(what_we_count) => model.what_we_count = what_we_count,
-//    }
-//    Render(model)
-//}
 
 // View
 
--- a/examples/counter/src/lib.rs
+++ b/examples/counter/src/lib.rs
@@ -63,7 +63,7 @@ fn success_level(clicks: i32) -> El<Msg> {
 
 /// The top-level component we pass to the virtual dom. Must accept the model as its
 /// only argument, and output a single El.
-fn view(_state: seed::App<Msg, Model>, model: &Model) -> El<Msg> {
+fn view(model: &Model) -> El<Msg> {
     let plural = if model.count == 1 { "" } else { "s" };
     let text = format!("{} {}{} so far", model.count, model.what_we_count, plural);
 
--- a/examples/counter/src/lib.rs
+++ b/examples/counter/src/lib.rs
@@ -35,15 +35,15 @@ fn update(msg: Msg, model: &mut Model) -> Update<Msg> {
     match msg {
         Msg::Increment => {
             model.count += 1;
-            Render
+            Render.into()
         }
         Msg::Decrement => {
             model.count -= 1;
-            Render
+            Render.into()
         }
         Msg::ChangeWWC(what_we_count) => {
             model.what_we_count = what_we_count;
-            Render
+            Render.into()
         }
     }
 }
--- a/examples/server_interaction/src/lib.rs
+++ b/examples/server_interaction/src/lib.rs
@@ -102,9 +102,9 @@ fn update(msg: Msg, model: &mut Model) -> Update<Msg> {
             Render.into()
         }
 
-        Msg::GetData => Update::with_future(get_data()).skip(),
+        Msg::GetData => Update::with_future_msg(get_data()).skip(),
 
-        Msg::Send => Update::with_future(send()).skip(),
+        Msg::Send => Update::with_future_msg(send()).skip(),
 
         Msg::OnServerResponse(result) => {
             log!(format!("Response: {:?}", result));
--- a/examples/server_interaction/src/lib.rs
+++ b/examples/server_interaction/src/lib.rs
@@ -44,16 +44,17 @@ struct Model {
     data: Branch,
 }
 
-fn get_data() -> impl Future<Item = Msg, Error = JsValue> {
+fn get_data() -> impl Future<Item = Msg, Error = Msg> {
    let url = "https://api.github.com/repos/david-oconnor/seed/branches/master";
 
    Request::new(url)
-       .method(Method::Get)
-       .fetch_json()
-       .map(Msg::Replace)
+        .method(Method::Get)
+        .fetch_json()
+        .map(Msg::Replace)
+        .map_err(Msg::OnFetchErr)
 }
 
-fn send() -> impl Future<Item = Msg, Error = JsValue> {
+fn send() -> impl Future<Item = Msg, Error = Msg> {
     let url = "https://infinitea.herokuapp.com/api/contact";
 
     let message = Message {
@@ -68,6 +69,7 @@ fn send() -> impl Future<Item = Msg, Error = JsValue> {
         .body_json(&message)
         .fetch_json()
         .map(Msg::OnServerResponse)
+        .map_err(Msg::OnFetchErr)
 }
 
 // Setup a default here, for initialization later.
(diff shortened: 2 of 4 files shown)


David-OConnor:
For the &mut model issue, can you accomplish what you need with the current api, and using the sig: `fn update(msg: Msg, model: mut Model) -> Update<Msg, Model>` ? This lets you mutate the model in place. Ideally, I'd like to keep the API so both in-place mutations, and returning a new model work.

Re triggering futures in updates: I think this is a great approach. Passing state down through the view functions, (or wrapped in messages) is clumsy, and is due to me struggling with implementations: It's not a deliberate API choice. I think you've solved it.

sapir:
Having the model parameter be mutable does allow it to be modified in-place, but I want to then pass part of it to a subcomponent function. i.e. My app structure looks something like this:

```rust
struct TextBoxModel {
    current_value: String,
}

fn update_textbox(msg: TextBoxMsg, mut model: TextBoxModel) -> Update<TextBoxMsg, TextBoxModel> {
    model.current_value = something;
    ...
    Render(whatever)
}

struct AppModel {
    text_box1: TextBoxModel,
    text_box2: TextBoxModel,
}

enum AppMsg {
    TextBox1Msg(TextBoxMsg),
    TextBox2Msg(TextBoxMsg),
}

fn update_app(msg: AppMsg, mut model: AppModel) -> Update<AppMsg, AppModel>
{
    ...
}
```

(This is of course a bit simplified: I have a few different component types, the real code has things split nicely into separate modules etc. etc.)

Now how do I implement update_app? The TextBox messages are supposed to be handled by update_textbox, which is supposed to get its TextBox model by value, then return it back to update_app wrapped inside an Update.

So I have a few options.

1. Split the AppModel into all its fields, so that the TextBox1 field is separate, pass it to update_textbox, get an `Update<TextBoxModel>` back, then add the rest of the fields back to get an `Update<AppModel>`. This is basically the way it works right now, and it looks something like this:

    ```rust
    fn update_app(msg: AppMsg, mut model: AppModel) -> Update<AppMsg, AppModel>
    {
        match msg {
            AppMsg::TextBox1Msg(textbox_msg) => {
                // Take the AppModel apart
                let AppModel { text_box1, text_box2 } = model;
                // Now we have text_box1 separate from the other fields. Call update_textbox.
                let text_box_update_object = update_textbox(textbox_msg, text_box1);
                // But now the update object only contains text_box1. We need to add
                // the rest of the AppModel.
                let app_update_object = match text_box_update_object {
                    Update::Render(text_box1) => Update::Render(
                        AppModel { text_box1, text_box2 },
                    ),

                    Update::Skip(text_box1) => Update::Skip(
                        AppModel { text_box1, text_box2 },
                    ),

                    Update::RenderThen(text_box1, text_box_msg) => {
                        let app_model = AppModel { text_box1, text_box2 };
                        let app_msg = AppMsg::TextBox1Msg(text_box_msg);
                        Update::RenderThen(app_model, app_msg)
                    }
                };
                // Now we can return it normally
                app_update_object
            }

            // The same thing all over again for text_box2.
            AppMsg::TextBox2Msg(textbox_msg) => { ... }
        }
    }
    ```

    There is a lot of DRY here. Looking at my example, it looks a bit like it's mostly inside the `match` expression, because I repeat the creation of the AppModel 3 times. But I can refactor that to a `wrap_update` function like I mentioned earlier.

    The real source of repetition is that I have to unpack *all* the AppModel fields to get ownership of text_box1, and then pack them *all* back again to get an AppModel. If I add a text_box3 field, I then have to add it in the code (twice) in the TextBox1Msg code_box1:

    ```rust
    let AppModel { text_box1, text_box2, text_box3 } = model;
    ...
    let new_appmodel = AppModel { text_box1, text_box2, text_box3 };
    ```

    And again for TextBox2Msg. So anyway, this option isn't so nice.

    I admit in the Elm architecture, `update` does return the model, but then again, Elm has no mutability at all, and instead all data modification is basically `clone`ing, only not as inefficient because of persistent data structures and other optimizations. So in Elm, I wouldn't have the problem I have here, because I would just be passing immutable references.

2. Use a mutable reference. As I understand it, this is the idiomatic Rust way of doing this kind of thing. Basically any function that takes a mutable reference could instead be taking ownership of the value, and then returning it ([like the book shows here](https://doc.rust-lang.org/stable/book/ch04-01-what-is-ownership.html#return-values-and-scope)), except that passing by value is a bit more repetitive, and also you run into the same problem I ran into here: You can't use by-value functions with parts of other structs/arrays/etc. because of the Rust ownership rules / borrow checker.

    Anyway, with a mutable reference, the code can be a lot simpler:

    ```rust
    fn update_app(msg: AppMsg, model: &mut AppModel) -> Update<AppMsg>
    {
        match msg {
            AppMsg::TextBox1Msg(textbox_msg) => {
                // We can call update_textbox directly now.
                let text_box_update_object = update_textbox(textbox_msg, &mut model.text_box1);
                // We still have to convert the message inside the Update. This can be
                // simplified further by helper functions and/or making Update a struct
                // and separating Update's Render/Skip field from the Option<Msg> part.
                let app_update_object = match text_box_update_object {
                    // The two look the same, but the left side is Update<TextBoxMsg>,
                    // but the right side is Update<AppMsg>
                    Update::Render => Update::Render,
                    Update::Skip => Update::Skip,
                    Update::RenderThen(text_box_msg) => {
                        let app_msg = AppMsg::TextBox1Msg(text_box_msg);
                        Update::RenderThen(app_msg)
                    }
                };
                // And now return it.
                app_update_object
            }

            // And again for text_box2.
            AppMsg::TextBox2Msg(textbox_msg) => { ... }
        }
    }
    ```

    The mutable reference means the function signature only has to mention AppModel once, but the real win is that the text_box1 code doesn't have to know about text_box2. And after refactoring away the message transformation, it's down to around 2 lines.

3. Use an Option to allow removing text_box1 from the AppModel before calling the inner update_textbox, and then put it back afterwards. This would allow passing everything by value, but now I have to have Options all over the place, and in general, it really isn't as nice as the previous option.

    ```rust
    // The struct changes now:
    struct AppModel {
        text_box1: Option<TextBoxModel>,
        text_box2: Option<TextBoxModel>,
    }

    fn update_app(msg: AppMsg, model: AppModel) -> Update<AppMsg, AppModel>
    {
        match msg {
            AppMsg::TextBox1Msg(textbox_msg) => {
                // Remove text_box1 from the App model.
                let text_box1 = model.text_box1.take().unwrap();
                // Pass it to update_textbox.
                let text_box_update_object = update_textbox(textbox_msg, text_box1);
                // Transform the Update object. This would be aided by a helper function,
                // and aided further by refactoring Update to a struct etc, but it'd still
                // be more complicated than it has to be.
                let app_update_object = match text_box_update_object {
                    Update::Render(text_box1) => {
                        model.text_box1.replace(text_box1);
                        Update::Render(model)
                    }

                    Update::Skip(text_box1) => {
                        model.text_box1.replace(text_box1);
                        Update::Skip(model)
                    }

                    Update::RenderThen(text_box1, text_box_msg) => {
                        let app_msg = AppMsg::TextBox1Msg(text_box_msg);
                        model.text_box1.replace(text_box1);
                        Update::RenderThen(model, app_msg)
                    }
                };
                // And now return it.
                app_update_object
            }

            // And again for text_box2.
            AppMsg::TextBox2Msg(textbox_msg) => { ... }
        }
    }
    ```

I'll see if I can separate this change from the rest of the PR. But I really would like to do option 2. What are the advantages to returning the model? Also, how would you do subcomponents? Is there some other way of structuring my app that I'm missing?

sapir:
I made a separate branch: https://github.com/David-OConnor/seed/compare/master...sapir:update-futures-no-mut-ref without the mutable reference change.

The other changes would still break the API, of course.

David-OConnor:
I'll dive in this weekend, and probably merge that one that has the update and futures changes. The futures change in particular looks like a big API improvement.

sapir:
Thanks

David-OConnor:
I don't have much to add, because of your detailed reasoning: I concur with all. This PR adds significant syntax improvements when the update logic is complex, and when using fetch requests.

PR #77, opened 2019-03-14
Support converting message types from other crates

hgzimmerman:
I think this is a pretty naive solution that wrapping the all the nested Listeners for the affected component function every time view() is called. If there was a lazy way to do this, so that the conversion only needs to happen when the listener is fired, this would be much lower overhead. 

I haven't tested this yet, but I imagine it could be used in some way like this:

```rust
struct Model {
    data: String,
    third_party: ThirdPartyModel
}

enum Msg {
    UpdateData,
    ThirdParty(ThirdPartyMsg)
}

impl From<ThirdPartyMsg> for Msg {
    fn from(other: ThirdPartyMsg) -> Self {
        Msg::ThirdParty(other)
    }
}

fn update(msg: Msg, &mut model: Model) -> Update<Msg> {
    match msg {
        ThirdParty(tp_msg) => third_party_update(tp_msg, &mut model.third_party),
        ...
    }
    ...
}

fn view(model: &Model) -> El<Msg> {
    div![
        third_party_view(&model.third_party).convert_message(),
        ...
    ]
}

### DIFF
--- a/src/dom_types.rs
+++ b/src/dom_types.rs
@@ -155,11 +155,6 @@ impl<Ms> fmt::Debug for Listener<Ms> {
     }
 }
 
-impl <Ms> Listener<Ms> 
-{
-
-}
-
 impl<Ms> Listener<Ms> {
     pub fn new(trigger: &str, handler: Option<EventHandler<Ms>>) -> Self {
         Self {
--- a/src/dom_types.rs
+++ b/src/dom_types.rs
@@ -155,6 +155,11 @@ impl<Ms> fmt::Debug for Listener<Ms> {
     }
 }
 
+impl <Ms> Listener<Ms> 
+{
+
+}
+
 impl<Ms> Listener<Ms> {
     pub fn new(trigger: &str, handler: Option<EventHandler<Ms>>) -> Self {
         Self {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/dom_types.rs [hunk 1]
```diff
@@ -1121,6 +1136,33 @@ impl<Ms> El<Ms> {
...
+    /// 
+    /// This allows the use of third party components to integrate with your application without
+    /// having to know about your Msg type beforehand. 
+    /// All that is required is that you implement `From<OtherMsg> for Msg` and hand off their
+    /// `OtherMsg` and section of state to their `update()` function (or equivalent) within your `update()` function.
+    ///
+    /// # Note
+    /// There is an overhead to calling this versus keeping all messages under one type.
+    /// The deeper the nested structure of children, the more time this will take to run.
+    pub fn convert_message<OtherMs: Into<Ms>>(other: El<OtherMs>) -> El<Ms> {
```
sapir: Maybe implement `From<El<OtherMs>> where OtherMs: Into<Ms>`?

hgzimmerman: I think that's what I initially tried, but ran afoul of the case where `OtherMs` is actually `Ms` (because `Ms` implements `Into<Ms>`), and `From<El<Ms>>` can't be reimplemented because the blanket `From<T> for T` impl already covers that case. I'll double-check this and get back to you.

sapir: Also, perhaps accept a callback to apply to the message, like I did for `Update.map`. This has the advantage that the caller can choose a different transformation, other than `OtherMs::into`. Also, maybe you can store the callback inside the `El` instead of applying it directly, to solve the overhead you mention in the comment.

hgzimmerman: I'll give that a shot.

I am sad to confirm that a From<El<OtherMs>> fails due to a conflicting implementation in core: `impl<T> std::convert::From<T> for T``.

hgzimmerman: I don't think storing the callback with `El` would be helpful, it A: necessitates that the other message type be part of the `El`, which would reduce ergonomics (eg: `El<Msg, OtherMsg>`) and still wouldn't scale beyond one external supported crate, and B: I think that the storage of the mapping function still needs to be done to all children recursively, which would still incur a similar cost.

David-OConnor:
I like the added functionality - merging despite the continued discussion; can submit a new PR later if needed. Holding off on updating the guide for now.

PR #79, opened 2019-03-15
Merge attrs

sapir:


David-OConnor:
I'd meant to do this one a while back. Held off due to trying to constrain the API, but I think the flexibility this provides wins out.

PR #83, opened 2019-03-19
Serve statics through Rocket in the server integration example

klemens:
If one is building an application using Seed and Rocket, it makes sense in my opinion to also serve the frontend through Rocket. This also avoids having to deal with CORS headers.

Feel free to ignore the second and third commit if you don't like them, they just improve the user experience a little bit.


David-OConnor:
I like all the changes. Might replace the build scripts in other examples with ones like this.

PR #84, opened 2019-03-21
Allow mounting elements

### LINKED ISSUE #40: Update AppBuilder mount method to accept Into<String>
Please update this code:

https://github.com/David-OConnor/seed/blob/5c30b6d354b9125b26f5b2ffe1187318022e67c2/src/vdom.rs#L86-L89

So that id accepts anything that implements `Into<String>`.

That way if users want to pass in an id that is dynamic, it is reasonably possible to do so while using the builder interface.

jgrund:
Fixes #40

This patch allows direct mounting of an element, which is useful
when you don't know the mount id ahead of time.

Signed-off-by: Joe Grund <jgrund@whamcloud.io>

### DIFF
--- a/src/vdom.rs
+++ b/src/vdom.rs
@@ -221,18 +221,29 @@ fn find_mount_point(id: &str) -> Element {
 }
 
 #[derive(Clone)]
+enum Parent {
+    Id(&'static str),
+    El(Element),
+}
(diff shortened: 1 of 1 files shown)


David-OConnor:
Good addition

PR #153, opened 2019-06-20
Implement MessageMapper for Listener

### LINKED ISSUE #152: implement MessageMapper for Listener
Currently `Listener` has its own private implementation for `map_message()`:
https://github.com/David-OConnor/seed/blob/632b134e7b74aa493a2ff9207c974da53321088a/src/dom_types.rs#L373

I think `Listener` should instead implement `MessageMapper`, also making it public so that I can remap subscriptions from any subcomponent modules

CallumJHays:
Fixes #152

### DIFF
--- a/src/dom_types.rs
+++ b/src/dom_types.rs
@@ -368,6 +368,25 @@ impl<Ms> Listener<Ms> {
     }
 }
 
+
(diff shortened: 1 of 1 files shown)


David-OConnor:
Looks good

PR #165, opened 2019-06-30
Child link listeners

David-OConnor:
https://github.com/David-OConnor/seed/issues/163

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -28,9 +28,6 @@ mod websys_bridge;
 
 // todo temporary: To allow `cargo publish` to work with the unreleased Gloo crate
 pub mod gloo_timers;
-//pub use gloo_timers;
-//pub use gloo_timers::callback::Timeout;
-//pub use gloo_timers::callback::Timeout;
 
 /// Create an element flagged in a way that it will not be rendered. Useful
 /// in ternary operations.
--- a/src/routing.rs
+++ b/src/routing.rs
@@ -1,6 +1,7 @@
 //! This module is decoupled / independent.
 
 use serde::{Deserialize, Serialize};
+use std::convert::identity;
 use wasm_bindgen::{closure::Closure, JsCast, JsValue};
 
 /// Repeated here from `seed::util`, to make this module standalone.  Once we have a Gloo module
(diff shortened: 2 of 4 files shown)


MartinKavik:
> // todo: Excessive nesting?

```rust
use std::convert::identity;

/// Set up a listener that intercepts clicks on elements containing an Href attribute,
/// so we can prevent page refresh for internal links, and route internally.  Run this on load.
#[allow(clippy::option_map_unit_fn)]
pub fn setup_link_listener<Ms>(update: impl Fn(Ms) + 'static, routes: fn(Url) -> Ms)
where
    Ms: 'static,
{
    let closure = Closure::wrap(Box::new(move |event: web_sys::Event| {
        event.target()
            .and_then(|et| et.dyn_into::<web_sys::Element>().ok())
            .and_then(|el| el.closest("[href]").ok())
            .and_then(identity)  // Option::flatten not stable (https://github.com/rust-lang/rust/issues/60258)
            .and_then(|href_el| match href_el.tag_name().as_str() {
                // Base and Link tags use href for something other than navigation.
                "Base" | "Link" => None,
                _ => Some(href_el)
            })
            .and_then(|href_el| href_el.get_attribute("href"))
            .and_then(|href| {
                href.chars().next()
                    .map(|first_char| (href, first_char))
            })
            // The first character being / indicates a rel link, which is what
            // we're intercepting.
            // todo: Handle other cases that imply a relative link.
            // todo: I think not having anything, eg no http/www implies
            // todo rel link as well.
            .and_then(|(href, first_char)| if first_char == '/' { Some(href) } else { None })
            .map(|href| {
                event.prevent_default(); // Prevent page refresh
                // Route internally based on href's value
                let url = push_route(Url::from(href));
                update(routes(url));
            });
    }) as Box<FnMut(web_sys::Event) + 'static>);

    (util::document().as_ref() as &web_sys::EventTarget)
        .add_event_listener_with_callback("click", closure.as_ref().unchecked_ref())
        .expect("Problem setting up link interceptor");

    closure.forget(); // todo: Can we store the closure somewhere to avoid using forget?
}
```

David-OConnor:
I don't like how repetitive that code chunk is, but not sure how to get around that; ie I think `else if` would still look weird.

MartinKavik:
I've tried some ideas, but I think it isn't possible without creating extra traits or some unsafe magic / reflection.
Merge at will, I'll try it and close the issue.

PR #196, opened 2019-08-12
Fixed a bug when rerendering multiple top-level nodes

David-OConnor:


src/websys_bridge.rs
```diff
@@ -155,20 +155,16 @@ pub fn attach_children<Ms>(el_vdom: &mut El<Ms>) {
...
 
     // Append the element
 
-    // todo: This can occur with raw html elements, but am unsur eof the cause.
-    match parent.append_child(el_ws) {
-        Ok(_) => {}
-        Err(_) => {
-            crate::log("Minor problem with html element (append)");
-        }
+    // todo: This error can occur with raw html elements, but am unsure of the cause.
```
MartinKavik: Could you create an issue with an example so that we don't forget it and I can try to debug it?

David-OConnor: I haven't encountered this in a while, and don't think it comes up anymore. I've replaced it locally with an expect call, and removed the todo. Will post here if it shows up.

PR #203, opened 2019-08-21
Added a release checklist

David-OConnor:


David-OConnor:
https://github.com/David-OConnor/seed/issues/137

MartinKavik:
Nice!

Tips:
- You can add something like _"Update Rust tools (e.g. `rustup update`)"_  before `5. Run cargo make verify...`

- I suggest to adding something like _"Check that CI pipeline passed"_ before `7. Run cargo package`.

PR #209, opened 2019-08-29
Add remaining SVG attributes to At:: macro

tiberiusferreira:
I copied the SVG attributes from https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute and ran 

```use std::fs::File;
use std::io;
use std::io::Read;
use inflector::Inflector;
fn filename_to_string(s: &str) -> io::Result<String> {
    let mut file = File::open(s)?;
    let mut s = String::new();
    file.read_to_string(&mut s)?;
    Ok(s)
}

fn main() {
    let file_as_str = filename_to_string("some.txt").unwrap();
    let mut i = 0;
    file_as_str.split_whitespace().for_each(|word| {
        print!("{} => \"{}\", ", word.to_pascal_case(), word);
        i = i + 1;
        if i == 3 {
            i = 0;
            println!();
        }
    })
}```
Then removed duplicates.

Hope it helps.

### CODE REVIEW

File: src/dom_types.rs [hunk 1]
```diff
@@ -273,7 +273,84 @@ make_attrs! {
...
+    RepeatDur => "repeatDur", RequiredExtensions => "requiredExtensions", RequiredFeatures => "requiredFeatures",
+    Restart => "restart", Result => "result", Rotate => "rotate",
+    Rx => "rx", Ry => "ry", Scale => "scale",
+    Seed => "seed", ShapeRendering => "shape-rendering", Slope => "slope",
+    Spacing => "spacing", SpecularConstant => "specularConstant", SpecularExponent => "specularExponent",
+    Speed => "speed", SpreadMethod => "spreadMethod", StartOffset => "startOffset",
+    StdDeviation => "stdDeviation", Stemh => "stemh", Stemv => "stemv",
+    StitchTiles => "stitchTiles", StopColor => "stop-color", StopOpacity => "stop-opacity",
+    StrikethroughPosition => "strikethrough-position", StrikethroughThickness => "strikethrough-thickness", String => "string",
+    Stroke => "stroke", StrokeDasharray => "stroke-dasharray", StrokeDashoffset => "stroke-dashoffset",
```
MartinKavik: `stroke-*` classes - StrokeDash**A**rray, StrokeDash**O**ffset...   I think it would be more readable and I think I will be making typos in them. Just idea, I haven't used them yet.

MartinKavik: > I agree with the readability point.

I think readability is the top priority for source code. So I vote for human-friendly version - we can always write better generators.

I'm looking again at those classes - we can also respect `-`s and replace them with `_` if it's more readable.

@David-OConnor What's your opinion?

David-OConnor: I see benefits in both perspectives, but agree with Martin - going straight CamelCase is consistent with Rust styling, which the rest of this lib adheres to. I think in cases like this, we should design the API the way we want it, without worrying about previous standards.

On the fence over straight CamelCase vs including both options...  API pureness vs flexibility. Eg we could include the `StrokeDasharray` ones, but not document or recommend using them. See later on if anyone relies on them.

David-OConnor: What are y'alls thoughts on having both available?

tiberiusferreira: I don't having strong feelings about this, but I prefer only having one option. It enforces better consistency and does not pollute IDEs autocomplete with different options of the same thing.

MartinKavik: I'm strongly against multiple versions. It would be very ugly in IDE's autocomplete and once somebody see them we can just wait for new issues like `what attribute we should use?`. And I don't want to think about it as the user.

David-OConnor: @tiberiusferreira  - If you update this PR with Martin's changes, we can merge this.

David-OConnor: Related: What do y'all think of having `St` enum: An equivalent to `At`, but for styles? The groundwork's been there in a WIP state for a while; ended up not continuing due to the [large number of possible styles](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference). I don't think it would hurt anything, but haven't heard a suggestion about this? The change would be to fill out [the make_styles!](https://github.com/David-OConnor/seed/blob/master/src/dom_types.rs#L340) macro call, add `St` to the prelude, and document it.

tiberiusferreira: I really like the idea, but I wonder if the increase in binary size will be negligible, after all the text will be stored (uncompressed before gzip, I imagine) somewhere in the binary. Anyway, I'd rather have the convenience even if at the cost of some KBs.

MartinKavik: Let's continue in https://github.com/David-OConnor/seed/issues/211
I think we can merge this PR now.

PR #212, opened 2019-09-02
Improve documentation for AppBuilder

spersson:
Needed to make the AppBuilder type public for its documentation to be
generated.

### DIFF
--- a/src/vdom.rs
+++ b/src/vdom.rs
@@ -203,19 +203,19 @@ impl<Ms, Mdl, ElC: View<Ms> + 'static, GMs: 'static> AppBuilder<Ms, Mdl, ElC, GM
         self
     }
 
-    /// Register a function which maps URLs to messages
+    /// Registers a function which maps URLs to messages.
     pub fn routes(mut self, routes: RoutesFn<Ms>) -> Self {
         self.routes = Some(routes);
         self
     }
 
-    /// Register a function which maps window events to messages
+    /// Registers a function which decides how window events will be handled.
     pub fn window_events(mut self, evts: WindowEvents<Ms, Mdl>) -> Self {
         self.window_events = Some(evts);
         self
     }
 
-    /// Register a sink function
+    /// Registers a sink function.
     ///
     /// The sink function is a function which can update the model based
     /// on global messages. Consider to use a sink function when a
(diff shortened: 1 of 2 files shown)


src/vdom.rs
```diff
@@ -203,21 +203,29 @@ impl<Ms, Mdl, ElC: View<Ms> + 'static, GMs: 'static> AppBuilder<Ms, Mdl, ElC, GM
...
+    ///
+    /// The sink function is a function which can update the model based
+    /// on global messages. Consider to use a sink function when a
+    /// submodule needs to trigger changes in other modules.
     pub fn sink(mut self, sink: SinkFn<Ms, Mdl, ElC, GMs>) -> Self {
         self.sink = Some(sink);
         self
     }
 
+    /// Turn this AppBuilder into an App which is ready to run.
```
MartinKavik: It would be nice to write comments according to [conventions](https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#appendix-a-full-conventions-text).

Could you add backticks to Rust entities (like `AppBuilder`, `sink`, etc.) or ideally links to the associated structs / methods? (It's possible that CI will fail on this)
And dots at the end of other comments?

MartinKavik:
Nice changes.

I've added one review comment and CI fails - try to run `cargo make verify` before the next push.
But the CI errors are a little bit weird, I'll try to fix it if your changes won't resolve it.

spersson:
Thanks.
I didn't run make verify because it takes me a long time to sit and click "approve" on a popup for every executable that the scripts wants to run. That's over a hundred times I think, a bit much since I didn't expect documentation changes to have any effect.
I tried improving the texts now, thank you for the link.

MartinKavik:
Nice! We can use these comments as the inspiration for future PRs.

---

> it takes me a long time to sit and click "approve" on a popup for every executable

Could you provide more context? (Or create a new issue if you think it's a bug?)

---

> I didn't run make verify

If it fails, I'll try to debug it tomorrow.

spersson:
Oh, the approve thing is just a security measure on my work laptop. It whitelists what is allowed to run. 
Yeah, let's see how that CI run goes, didn't look related to these changes. Would appreciate the help!

MartinKavik:
Everything seems good to me, I think we are ready for merge. 
cc @David-OConnor

PR #213, opened 2019-09-02
Style enums

David-OConnor:
Not tested, and needs a way to to execute, eg makefile, or turn into a workspace crate.

https://github.com/David-OConnor/seed/issues/211


David-OConnor:
Do you want to add anything else before merging? There's still applying this to events/tags/attrs, and dealing with allowed values, but as-is, this provides a nice API improvement.

MartinKavik:
Yeah, just one task (https://github.com/David-OConnor/seed/pull/216) so we are ready for other entities.

PR #221, opened 2019-09-19
Add escape hatch from custom routing.

AlterionX:
I was attempting to load in `seed` from an endpoint other than `/`. After a while struggling with attempting to manually route using `web-sys`, I decided to simply provide an escape hatch by allowing routing to return an `Option`.

I hope that the usage is relatively intuitive, but if you return a `None`, routing will fail and fall back to the default operation. If a `Some` is returned, then routing will happen as per the original routing code.

Maybe this should be done a different way? But this was the simplest solution I could think of.

### DIFF
--- a/src/routing.rs
+++ b/src/routing.rs
@@ -215,7 +215,7 @@ pub fn push_route<U: Into<Url>>(url: U) -> Url {
 pub fn setup_popstate_listener<Ms>(
     update: impl Fn(Ms) + 'static,
     update_ps_listener: impl Fn(Closure<dyn FnMut(web_sys::Event)>) + 'static,
-    routes: fn(Url) -> Ms,
+    routes: fn(Url) -> Option<Ms>,
 ) where
     Ms: 'static,
 {
@@ -234,7 +234,10 @@ pub fn setup_popstate_listener<Ms>(
                 Url::new(empty)
             }
         };
-        update(routes(url));
+        // Only update when requested for an update by the user.
+        if let Some(routing_msg) = routes(url) {
+            update(routing_msg);
+        }
     });
 
     (util::window().as_ref() as &web_sys::EventTarget)
@@ -247,7 +250,7 @@ pub fn setup_popstate_listener<Ms>(
 /// Set up a listener that intercepts clicks on elements containing an Href attribute,
 /// so we can prevent page refresh for internal links, and route internally.  Run this on load.
 #[allow(clippy::option_map_unit_fn)]
-pub fn setup_link_listener<Ms>(update: impl Fn(Ms) + 'static, routes: fn(Url) -> Ms)
+pub fn setup_link_listener<Ms>(update: impl Fn(Ms) + 'static, routes: fn(Url) -> Option<Ms>)
 where
     Ms: 'static,
 {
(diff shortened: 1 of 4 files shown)


src/routing.rs
```diff
@@ -274,12 +276,16 @@ where
...
             })
             .map(|href| {
-                event.prevent_default(); // Prevent page refresh
                 // @TODO should be empty href ignored?
-                if !href.is_empty() {
+                if href.is_empty() {
+                    event.prevent_default(); // Prevent page refresh
+                } else {
                     // Route internally based on href's value
                     let url = push_route(Url::from(href));
```
MartinKavik: Why is `push_route` outside of the condition below?
(Why not something like this:
```rust
let url = Url::from(href);
if let Some(redirect_msg) = routes(url) {
    push_route(url);
    event.prevent_default(); // Prevent page refresh
    update(redirect_msg);
}
```
?)
I don't know History API so well, I'm just curious.

MartinKavik:
I think it's a good idea, I was thinking about something similar a few days ago - see https://github.com/MartinKavik/kavik.cz/blob/master/crate/src/route.rs#L16 - it's a link to `pdf` file. Seed doesn't recognize files as external links so I had to write the ugly workaround which redirects browser. But even then, the history stack seems to be a little bit broken.

P.S. Travis job probably failed because of some network issues (I have similar problems in repo `html-css-db`. Restart should help.

src/routing.rs
AlterionX: Hmmm, wasn't really thinking about it when I added the change., I was just trying to keep behavior as consistent as possible when modifying the code.

AlterionX:
Note taken on the Travis.

Seems like it's still being a bit wonky on the whole history thing (routing that hits the "routes" closure and nothing happens doesn't seem to want to obey the prev and next page buttons (at least on chrome)), so I'm still looking into it.

Hopefully I can fix that soon and put that down as part of this PR as well.

src/routing.rs
AlterionX: Wait, I totally misremembered my code. I will look into it. I think it was because I wasn't thinking about history. Ooops.

David-OConnor:
There's also a problem with landing-page routing I didn't notice before. Will take a look.

src/routing.rs
AlterionX: You were right about `push_route`! The history buttons finally work again.

MartinKavik:
> Hopefully I can fix that soon and put that down as part of this PR as well.

Don't hurry, it's not easy to get it right. 
I've added link to Elm routing implementation when I was fixing something in Seed's routing - you can use it as inspiration: https://github.com/David-OConnor/seed/blob/master/src/routing.rs#L268

AlterionX:
Hmmm, I ended up having to add a `.clone()` to get around the fact that the `routes` function consumes the url.

MartinKavik:
> add a .clone() to get around the fact that the routes function consumes the url.

This is the reason why `routes` consumes the url: https://github.com/David-OConnor/seed/issues/139#issue-456446917. Probably still valid, but maybe you'll find more elegant solution and we can use & after that.

However if you want to optimize something I would rather focus on the extra `clean_url` call (it's inside `push_route`) than on the `clone`. But I think both "problems" doesn't really matter - _"Premature optimization is the root of all evil"_. So I suggest to resolve original issue with this code - I didn't test it but it looks good and we can create a new issue like _"Implement routing properly with tests and optimizations"_ later.

AlterionX:
Well, I guess it works for now. Would someone mind merging this?

David-OConnor:
Nice change. Worth the break.

PR #227, opened 2019-09-24
Init routing

David-OConnor:
https://github.com/David-OConnor/seed/issues/224

Tests currently broken.

### DIFF
--- a/src/vdom.rs
+++ b/src/vdom.rs
@@ -67,7 +67,7 @@ type MsgListeners<Ms> = Vec<Box<dyn Fn(&Ms)>>;
 /// Used for handling initial routing.
 pub enum UrlHandling {
     PassToRoutes,
-    Nothing,
+    None,
     // todo: Expand later, as-required
 }
 
@@ -285,7 +285,7 @@ impl<Ms, Mdl, ElC: View<Ms> + 'static, GMs: 'static> AppBuilder<Ms, Mdl, ElC, GM
                     (self.update)(r(url).unwrap(), &mut init.model, &mut initial_orders);
                 }
             }
-            UrlHandling::Nothing => (),
+            UrlHandling::None => (),
         };
 
         app.cfg.initial_orders.replace(Some(initial_orders));
--- a/src/vdom.rs
+++ b/src/vdom.rs
@@ -1425,10 +1425,12 @@ pub mod tests {
         }
 
         let app = App::build(
-            |_, _| Init::new(Model {
-                test_value_sender: Some(test_value_sender),
-                ..Default::default()
-            }),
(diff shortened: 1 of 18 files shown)


MartinKavik:
WIP, but look good! 
Could you notify me before merge? I want to try some type magic - e.g. replace `-> Init<Model>` with `-> impl Init`. I don't know if it's doable or better but I want to try it and eventually compare both versions.

David-OConnor:
CI issue was due to fmt. Have at the `impl` - that'd be more consistent with `View` and `Update`.

MartinKavik:
How would look like standalone `init`?
i.e. instead of
```rust
fn init(_: Url, orders: &mut impl Orders<Msg>) -> Init<Model> {
    orders.perform_cmd(fetch_repository_info());
    Init::new(Model::default())
}
```
would be..?

David-OConnor:
I was thinking it could be a closure as a closure, but on reflection, that may not be robust enough. eg

```rust
let init = Init::from_fn(|_, | {
    orders.perform_cmd(fetch_repository_info());
    Init::new(Model::default())
})
```

MartinKavik:
And.. what about:
i.e. standalone init (without changes)
```rust
fn init(_: Url, _: &mut impl Orders<Msg>) -> Init<Model> {
    Init::new(Model::default())
}
```
One-liner:
```rust
seed::App::build(Init::default_fn(Model::default()), update, view)
```

src/vdom.rs
```diff
@@ -53,15 +53,47 @@ pub enum ShouldRender {
...
 type SinkFn<Ms, Mdl, ElC, GMs> = fn(GMs, &mut Mdl, &mut OrdersContainer<Ms, Mdl, ElC, GMs>);
 type ViewFn<Mdl, ElC> = fn(&Mdl) -> ElC;
 type RoutesFn<Ms> = fn(routing::Url) -> Option<Ms>;
 type WindowEvents<Ms, Mdl> = fn(&Mdl) -> Vec<events::Listener<Ms>>;
 type MsgListeners<Ms> = Vec<Box<dyn Fn(&Ms)>>;
 
+/// Used for handling initial routing.
+pub enum UrlHandling {
+    PassToRoutes,
+    Nothing,
```
MartinKavik: Why not `None`?

David-OConnor: Agree

David-OConnor:
Want to try `impl Init`? Ready to merge as-is, pending fixing CI.

MartinKavik:
I'm not able to come up with better API. I see some comments in PR which we can delete, but we can remove them with some future refactors. Merge it.

PR #233, opened 2019-10-05
Factor out duplicate logic in `vdom::rerender_vdom` and `path::patch_el`

AlterionX:
Resolves a bug, where if the top level `view` function returns a `Vec` of `Node`s and you navigate away to a different page with a different number of top level elements, weird things happen.

Weird things meaning that routing and rerendering/diffing the vdom adds extra elements to the mount point.


David-OConnor:
Very nice. In addition to fixing the bug, that duplicate chunk of code you factored out has been a maintanability-thorn ever since adding the ability to return a top-level `vec`. It's caused a number of rendering bugs we applied bandaid-fixes to.

Issue #232, opened 2019-10-04
Question/Feature suggestion: Server side rendering


AlterionX:
Is server side rendering being considered at all?

It seems borderline impossible in my head without severe restrictions, but I was just wondering.

MartinKavik:
Why do you need it?

I use Seed as a static site generator as alternative. See [seed-quickstart-webpack](https://github.com/MartinKavik/seed-quickstart-webpack) for example.

AlterionX:
I've got a login system that dynamically changes the nav menu. I think I want this to happen on the server, since if someone navigates to my site, they'll see the nav menu flash.

To do this, I think I need some form of SSR? Then I got carried away thinking about everything else that could use it.

Also, the rest of my static pages uses `maud` for rendering, so then I'd need to combine those two systems.

Totally missed the quickstart thing, btw, so thanks.

MartinKavik:
> nav menu flash

I want to fix one type of flashes - see issue https://github.com/David-OConnor/seed/issues/223.

But you probably think the flash of content once you loaded info about user and want to show e.g. his name and avatar in the menu. I think there are at least two options:

  1. If the site is NOT prerendered, you can load required data in `init` function (ideally from `localStorage` - we want it to be synchronous and fast). E.g. [seed-rs-realworld](https://github.com/MartinKavik/seed-rs-realworld/blob/c231f8a6f2ded4f555023da204d9e76ac39c1c0d/src/lib.rs#L69) uses `localStorage` that way.

  2. If the site IS prerendered, you can load required data also in `init` function, but you don't want to block first render. So you can show some placeholders or spinner instead of dynamic page parts.
     - Example of placeholders: [leerob.io/blog/loading-placeholder-with-sass](https://leerob.io/blog/loading-placeholder-with-sass).
     - Seed example with spinner: [kavik.cz](https://kavik.cz/) - change the window width to see hamburger and enable throttling in dev tools. Don't do the same mistake like me - inline you images or otherwise your icons can disappear/show too late during loading. Code is [here](https://github.com/MartinKavik/kavik.cz/blob/d16de781bcd79be3f96d493d682e04227df5b696/crate/src/page/partial/header.rs#L453).

> Also, the rest of my static pages uses maud for rendering, so then I'd need to combine those two systems.

Try to render `maud` templates into `String` in `init` and then render the result in `view` function with macro `raw!í..)`. Prerender your app and you get HTML static files.

AlterionX:
[...]

Issue #223, opened 2019-09-19
Proposal: Hydrate


MartinKavik:
## Motivation
Flashing page elements on wasm file load on prerendered pages.

## How to reproduce
Try to open [kavik.cz/about/](https://kavik.cz/about/) on a mobile or in a desktop browser with enabled throttling in dev tools. The first you see prerendered html and once the wasm is loaded, the entire page is rerendered which causes flashes.

## Problem
This code: https://github.com/MartinKavik/kavik.cz/blob/master/crate/src/app.rs#L49. I have to delete root element content before render, otherwise Seed will append identical html at the end of the root.

## How to resolve
Use prerendered HTML as the previously rendered HTML by Seed. So either introduce a new function `hydrate` or make `render` more intelligent (probably better option). There are examples from other frameworks: https://github.com/stereobooster/react-snap#basic-usage-with-create-react-app.

## Possible implementation problems
Prerendered and rendered code isn't 100% identical, e.g. hamburger is spinner in prerendered version (see my example in section `how to reproduce`).

## Public API changes.
None, if we update `render` function.

PR #235, opened 2019-10-06
Change vdom bootstrapping to occur before the initial orders are processed.

AlterionX:
# Changes:
- Change execution of the initial call to `process_cmd_and_msg_queue` to after the `main_el_vdom` is initialized. This allows for the msg processing to not cause issues if it attempts to force a render or something else similar. Forcing a render on init used to cause Seed to panic.
- Allow Seed to take over the children of the mount. This can be useful for SSR. Built-in SSR for Seed is my goal for now. This is partially related to #232, which is also related to #223.
- Warns of possibly questionable behavior when attempting to accomplish the above maneuver as it is now (see the doc comment for `takeover_mount` in `AppBuilder`). Due to this behavior and backwards compatibility, it is currently an opt in functionality.
- Adds a bunch of utility methods to the `Node` enum.
- Adds a function to recursively remove all workspace `web_sys::Node`s.
# Parts that need additional consideration (incomplete)
- Unsure if the listeners in the bootstrap should be set up -- not familiar with what they do. But it seems to work for now.
- Unsure of any other implications of moving the initial order processing to after the initialization of the `main_el_vdom`.
- Not sure if any of the utility methods added to the `Node` enum are actually of use.
- Might want to fully address the "recreation" behavior of `bootstrap_vdom` before this gets merged.

### DIFF
--- a/src/dom_types.rs
+++ b/src/dom_types.rs
@@ -749,8 +749,8 @@ impl<Ms: Clone> Node<Ms> {
 
     /// See `El::add_child`
     pub fn add_child(&mut self, node: Node<Ms>) -> &mut Self {
-        if let Node::Element(el) = self {
-            el.add_child(node);
+        if let Node::Element(e) = self {
+            e.add_child(node);
         }
         self
     }
@@ -761,55 +761,70 @@ impl<Ms: Clone> Node<Ms> {
         key: impl Into<Cow<'static, str>>,
         val: impl Into<AtValue>,
     ) -> &mut Self {
-        self.map_el_in_place(move |el| {
-            el.add_attr(key, val);
-        })
+        if let Node::Element(e) = self {
+            e.add_attr(key, val);
+        }
+        self
     }
 
     /// See `El::add_class``
     pub fn add_class(&mut self, name: impl Into<Cow<'static, str>>) -> &mut Self {
-        self.map_el_in_place(move |el| {
-            el.add_class(name);
-        })
+        if let Node::Element(e) = self {
+            e.add_class(name);
+        }
+        self
     }
 
     /// See `El::add_style`
     pub fn add_style(&mut self, key: impl Into<St>, val: impl Into<CSSValue>) -> &mut Self {
-        self.map_el_in_place(move |el| {
-            el.add_style(key, val);
-        })
+        if let Node::Element(e) = self {
+            e.add_style(key, val);
+        }
+        self
     }
 
     /// See `El::add_listener`
     pub fn add_listener(&mut self, listener: Listener<Ms>) -> &mut Self {
-        self.map_el_in_place(move |el| {
-            el.add_listener(listener);
-        })
+        if let Node::Element(e) = self {
+            e.add_listener(listener);
+        }
+        self
     }
 
     /// See `El::add_text`
     pub fn add_text(&mut self, text: impl Into<Cow<'static, str>>) -> &mut Self {
-        self.map_el_in_place(move |el| {
-            el.add_text(text);
-        })
+        if let Node::Element(e) = self {
+            e.add_text(text);
+        }
+        self
     }
 
     /// See `El::replace_text`
     pub fn replace_text(&mut self, text: impl Into<Cow<'static, str>>) -> &mut Self {
-        self.map_el_in_place(move |el| {
-            el.replace_text(text);
-        })
+        if let Node::Element(e) = self {
+            e.replace_text(text);
+        }
+        self
     }
 
     /// See `El::get_text`
     pub fn get_text(&self) -> String {
-        self.map_or_else_ref(El::get_text, Text::get_text, "".to_owned())
+        match self {
+            Node::Element(e) => e.get_text(),
+            Node::Text(t) => t.get_text(),
+            _ => "".to_owned(),
+        }
     }
 }
 // Workspace manipulations.
 impl<Ms: Clone> Node<Ms> {
     pub(crate) fn strip_ws_nodes(&mut self) -> &mut Self {
-        self.map_in_place(El::strip_ws_nodes, Text::strip_ws_nodes)
+        match self {
+            Node::Element(e) => e.strip_ws_nodes(),
+            Node::Text(t) => t.strip_ws_nodes(),
+            _ => (),
+        }
+        self
     }
 }
 // Convenience methods for `Node` a la `Result` and `Option`.
@@ -864,36 +879,6 @@ impl<Ms: Clone> Node<Ms> {
         match self {
             Node::Empty => Ok(()),
             node => Err(node),
-        }
-    }
-
-    fn map_in_place<TMap, EMap>(&mut self, el_map: EMap, text_map: TMap) -> &mut Self
-    where
-        TMap: FnOnce(&mut Text),
-        EMap: FnOnce(&mut El<Ms>),
-    {
-        match self {
-            Self::Element(el) => el_map(el),
-            Self::Text(text) => text_map(text),
-            Self::Empty => (),
-        }
-        self
-    }
-    fn map_el_in_place<EMap>(&mut self, el_map: EMap) -> &mut Self
-    where
-        EMap: FnOnce(&mut El<Ms>),
-    {
-        self.map_in_place(el_map, |_| ())
-    }
-    fn map_or_else_ref<T, TMap, EMap>(&self, el_map: EMap, text_map: TMap, default: T) -> T
-    where
-        TMap: FnOnce(&Text) -> T,
-        EMap: FnOnce(&El<Ms>) -> T,
-    {
-        match self {
-            Self::Element(el) => el_map(el),
-            Self::Text(text) => text_map(text),
-            _ => default,
         }
     }
 }
--- a/src/dom_types.rs
+++ b/src/dom_types.rs
@@ -717,6 +717,12 @@ impl Text {
             node_ws: None,
         }
     }
+    pub fn get_text(&self) -> String {
+        self.text.to_string()
+    }
+    pub fn strip_ws_nodes(&mut self) {
+        self.node_ws.take();
+    }
 }
 
 /// An component in our virtual DOM. Related to, but different from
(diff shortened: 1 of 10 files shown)


MartinKavik:
@AlterionX Your `changes` list looks promising! I'll try to do code review and think about `incomplete` list in a few days.

MartinKavik:
@AlterionX 
- It's possible that this PR also resolves https://github.com/David-OConnor/seed/issues/198 
- I cloned your `master` (aka this PR) and when I start an example (e.g. `cargo make start counter`) there is no errors but also no content (blank page). So I assume you are working on it and I don't want to change code under your hands => let me know when you need some help / it's ready for some (partial) review. Thanks.

AlterionX:
Oops, I forgot to force a render at the start. I was primarily doing testing with my own app, so I'll try to fix that now.

MartinKavik:
So I've done a first round of code review:

1. We should try to implement traits instead of writing new custom methods. E.g.:
    - `src/dom_types.rs` - `Text::get_text(..)` can be replaced with `impl Display for Text`
    - `src/websys_bridge` -  `el_from_ws_element(..)` => `impl From<&web_sys::Element> for El<Ms>`

1. Match function name with it's body.
   - E.g. method `strip_ws_nodes`:
        - Why the body is `self.node_ws.take();` and not `self.node_ws = None;` => should we rename it to `take_ws_node(s)`?
        - Why the name is ending with `s`? `Text` has only one `node_ws`. If it should work the same for more entities (e.g. `El`, `Svg`, `Text`), we should abstract it to a trait or implement it directly on `Node` (if it makes sense).
        - It's a little bit surprising that `strip_ws_nodes` deletes also children nodes for `El` - at least a comment would be nice.

1. Probably unnecessary helpers like `map_el_in_place`, `map_or_else_ref`, `map_in_place`, `strip_ws_nodes` and `process_cmd_and_msg_queue_with_forced_render`.
   - Is it code really more readable and maintanable with them?
   - It's possible that there will be `Node` variant `Svg`. We'll have to update `map_in_place`, `map_or_else_ref`, many methods like `El:add_attr(..)`, write method `map_svg_in_place`, etc. - is it a good trade-off?

1. Don't be afraid to extract some parts of code into standalone modules (files) - see `/src/dom_types/values.rs`. E.g.:
    - `el_from_ws_element` + `node_from_ws` from `src/websys_bridge.rs`
    - `AppBuilder` + its implementations / helpers from `vdom.rs`

1. Don't be afraid to extract some parts of big functions.
    - E.g. `el_from_ws_element` - block with comment `Populate attributes` and the last block with the loop. (Comments like `Populate attributes` often signal that we should extract code below to functions with the similar name).

1. More flexible / simpler public API for `AppBuilder::takeover_mount(..)`.
    - I think we don't need parameter `should_takeover` - we just don't call that method at all.
    - What if we need another rendering mode (e.g. we'll find out that we need different modes for prerendering and SSR)? I'll basically don't like `bool`s because they aren't expressive enough and you can't add more values into them in the future. 
        - I suggest to introduce a new enum like `MountType` = `Takeover` | `Append`.
        - And `.takeover_mount()` replace with `.mount_type(MountType::TakeOver)`.
        - Another option is to reuse `.mount` but it will be breaking change and you would have to define explicitly mount point even if you want to use the default one.
       - cc @David-OConnor 

---

I didn't dive deeper into code logic (e.g. why `Unrecognized tags are also converted into spans, so take note.`). I'll do it in the next round.

@AlterionX Well, it looks like I don't like the code but it's not true, I think it's very good and resolves many pains - my points are opinionated and the PR has many changes so it's a good opportunity to refactor it a little bit during process.

AlterionX:
All good points. I didn't even really think about it.

Regarding `get_text`, that was simply a method moved from elsewhere. Most of these changes simply followed what existed before.

Not sure if I should here, but I can try scanning over the entire library to see if there are some other things that can be pulled out?

MartinKavik:
> Regarding get_text, that was simply a method moved from elsewhere. 

There are parts which doesn't make sense now or isn't super clean - don't be afraid to change it / create issue / write todo. (Just like every other project basically.)

> Try scanning over the entire library to see if there are some other things that can be pulled out?

I recommend to focus mostly on things which you are changing or are somehow related - just make it not too big for review and resolving conflicts with other PRs :)

AlterionX:
Got a question: is there a reason for the functions being put into the model being function pointers instead of `Fn` trait structs? If not, I'm thinking of changing that in a different PR.

MartinKavik:
> is there a reason for the functions being put into the model being function pointers

I'm not sure which functions do you mean (?). But in general - we try to write every public API with `FnOnce` for maximum flexibility (https://github.com/David-OConnor/seed/issues/169#issuecomment-515673391).

AlterionX:
I'm referring to `SinkFn`, `UpdateFn`, etc. Also, I think I figured out why they're that way, so nevermind.

MartinKavik:
I think that there aren't any specific reasons. Why are you asking?

AlterionX:
I think if you make it generic, the compiler will complain that it can't figure out what the functions are actually supposed to be if you don't use, say, `sink`, when building the `AppBuilder`. This would then force you to specify the full type of `AppBuilder`, which is long as heck, making it extremely not ergonomic for the base case where you only use build. This is mostly just speculation, since I haven't actually tried it out, but still.

MartinKavik:
You can write them as `Fn`, I just don't see any real-world advantage now. Create PR if you find it useful.
E.g. `SinkFn` can rewritten to:
```rust
type SinkFn<Ms, Mdl, ElC, GMs> = Box<dyn Fn(GMs, &mut Mdl, &mut OrdersContainer<Ms, Mdl, ElC, GMs>)>;
```
(and compiled with one small change in code).

AlterionX:
Hrm, not exactly what I meant.

I was wondering why `SinkFn` wasn't something like

```rust
trait SinkFn<Ms, Mdl, ElC, GMs>: FnMut(GMs, &mut Mdl, &mut OrdersContainer<Ms, Mdl, ElC, GMs>) {}
impl<Ms, Mdl, ElC, GMs, F> SinkFn for F
where
    F: FnMut(GMs, &mut Mdl, &mut OrdersContainer<Ms, Mdl, ElC, GMs>) {}
```

AlterionX:
@MartinKavik @David-OConnor I'm starting to get half a mind to move the `Init` from `build` to `run` or something similar to isolate all the side effects inside of `run` instead of distributing the stateful effects across `finish` and `run`.

I think that this is better since we do this weird dance around `initial_orders` right now where we don't quite have the correct state to do what we have half the state we want inside `Builder` and half the state inside the `App`.

So we would have
```rust
App::build(update, view).finish().run_with_init(|_, _| {
    // Do things with url + orders.
    // `Init::default` is a wrapper for `Init::new(Model::default())`.
    Init::default()
})
```
or
```rust
App::build(update, view).finish().run() // Calls `Init::default` by default.
```

It's a bit of a departure from what we currently have, but I think that it makes the implementation a little less weird and the usage clearer about when exactly the `InitFn` happens (since it seems to run half in `finish` and half in `run` right now).

The downsides, I think, are also fairly obvious. Also, the model is now provided in the `run` method instead of the `build` method, which is also... unique, so to speak.

AlterionX:
Whoops, didn't mean to do that.

AlterionX:
> Why the body is `self.node_ws.take();` and not `self.node_ws = None;` => should we rename it to take_ws_node(s)?

`self.node_ws.take();` and `self.node_ws = None;` accomplish the same goal, so I don't think it matters. The end result is still the same. If we want to get really banal about it, we could say that `take` leaves the enum in place, while `= None` overwrites the entire enum, but it's not that important.

> I think we don't need parameter should_takeover - we just don't call that method at all.

Not entirely sure what method this is talking about. `should_takeover` is a field, and is passed on to the app, which uses it later.

AlterionX:
Closing to trigger another travis build -- I think it might be a Mac specific error, but it seemed to work just fine on linux.

David-OConnor:
@MartinKavik : Any objections to merging this?

src/dom_types.rs
```diff
@@ -766,63 +770,129 @@ impl<Ms: Clone> Node<Ms> {
...
+    }
+    pub fn is_empty(&self) -> bool {
+        if let Node::Empty = self {
+            true
+        } else {
+            false
+        }
+    }
+
+    pub fn el(self) -> Result<El<Ms>, Option<Text>> {
```
MartinKavik: Return value seems a little bit weird to me, I would expect something like:
```rust
pub fn el(self) -> Option<El<Ms> { .. }
```
or
```rust
pub fn el(self) -> Result<El<Ms>, String> { .. }
pub fn el(self) -> Result<El<Ms>, ()> { .. }
```

---

The same comment for methods `text` and `empty`.

src/vdom.rs
```diff
@@ -370,49 +199,101 @@ impl<Ms: Clone, Mdl, ElC: View<Ms> + 'static, GMs: 'static> App<Ms, Mdl, ElC, GM
...
-        new.children = (self.cfg.view)(self.data.model.borrow().as_ref().unwrap()).els();
+    /// Bootstrap the dom with the vdom by taking over all children of the mount point and
+    /// replacing them with the vdom if requested. Will otherwise ignore the original children of
+    /// the mount point.
+    fn bootstrap_vdom(&self) -> El<Ms> {
+        let bootstrap_behavior = *self.cfg.bootstrap_behavior.borrow();
+        let mut new = El::empty(dom_types::Tag::Placeholder);
+
+        // Map the DOM's elements onto the virtual DOM if requested to takeover.
+        if bootstrap_behavior == Some(BootstrapBehavior::Takeover) {
```
MartinKavik: Can we rename it to `mount_type` and `MountType` so users don't have to relearn name once we change API to this one: https://github.com/David-OConnor/seed/issues/250#issuecomment-544512869?

MartinKavik:
@David-OConnor I wrote some comments. It seems that the only public API change is `mount_type` (`bootstrap_behavior`) in `Init` (?). If examples work and once the comments are resolved, I'm ok with merge.

AlterionX:
I'm trying to pull out a few things from this PR into smaller PRs. Will fix the names and return types soon.

src/dom_types.rs
AlterionX: Changed in a different PR (#271).

AlterionX:
Closing since changes are implemented in #269, #270, #271, #274, #276.

PR #239, opened 2019-10-11
Fixed a bug with svg children

David-OConnor:
Fixes https://github.com/David-OConnor/seed/issues/236

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## v0.4.2
 - Added an `Init` struct, which can help with initial routing (Breaking)
+- Fixed a bug with children being absent from cloned elements
+- Improved debugging
+- Fixed a namespace bug with adding children to `Svg` elements
 
 ## v0.4.1
 - Added more SVG `At` variants
--- a/src/dom_types.rs
+++ b/src/dom_types.rs
@@ -130,26 +130,29 @@ impl<Ms: Clone> UpdateEl<El<Ms>> for &str {
 
 impl<Ms: Clone> UpdateEl<El<Ms>> for El<Ms> {
     fn update(self, el: &mut El<Ms>) {
-        el.children.push(Node::Element(self))
+        el.add_child(Node::Element(self));
     }
 }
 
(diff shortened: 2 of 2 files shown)


MartinKavik:
I have some concerns:
1.  `children` is `pub` - I'm not sure that fixing `namespace` in `add_child` is a good idea. Once you use `children` directly (in Seed core or in apps), the problem is back.

1. SVG elements can contain non-SVG elements, see https://developer.mozilla.org/en-US/docs/Web/SVG/Element/foreignObject. It's possible that HTML elements in SVG element must have `namespace` set to `xmlns="http://www.w3.org/1999/xhtml` - it makes it even more difficult / error-prone. 

1. I also think that changing element in `add_child` is very surprising.
We should at least make it more transparent, e.g.:
```rust
/// Add a new child to the element.
///
/// _Note_: Children inherit the `Svg` namespace from parents.
pub fn add_child(&mut self, mut element: Node<Ms>) -> &mut Self {
    // Let children inherit the `Svg` namespace from parents.
    if self.namespace == Some(Namespace::Svg) {
     ....
```

---

I like your solution - it's elegant and simple, but I'm afraid that this issue is more complicated.
I recommend to find out more (edge-)cases like `foreignObject` and write some tests, TDD would be probably nice for this issue.

David-OConnor:
Alternatively, we could leave the code the same, and recommend users apply the namespace manually when adding children the way @ MuhannadAlrusayni did.

MartinKavik:
It's painful imho. Are we able to determine if the element is HTML or SVG during its creation - e.g. from `tag`? If so we can at least set reasonable default `namespace`.

David-OConnor:
Yes - that might be a more direct and side-effect-free approach. The `svg`, `rect` etc macros know to set the SVG namespace, but `raw_html` doesn't. This issue appears to be specific to adding a child svg as `raw_html` that doesn't specify namespace.

MartinKavik:
> The raw_html parsing is pretty well.. raw, but perhaps there's a way to intercept and inspect the tags.

Good luck, if you find some problems, let me know.

David-OConnor:
https://github.com/David-OConnor/seed/pull/240

PR #240, opened 2019-10-11
Intercept raw html SVG elements, and set their namespace

David-OConnor:
https://github.com/David-OConnor/seed/issues/236, take 2

Need to figure out how to populate the tags to intercept. Could just dump them in with the current, or somehow use the macro tags.

### DIFF
--- a/src/websys_bridge.rs
+++ b/src/websys_bridge.rs
@@ -1,6 +1,6 @@
 //! This file contains interactions with `web_sys`.
 use crate::dom_types;
-use crate::dom_types::{AtValue, El, Node, Text, Namespace};
+use crate::dom_types::{AtValue, El, Namespace, Node, Text};
 
 use wasm_bindgen::JsCast;
 use web_sys::Document;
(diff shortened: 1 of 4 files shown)


src/websys_bridge.rs
```diff
@@ -364,6 +364,16 @@ pub fn node_from_ws<Ms: Clone>(node: &web_sys::Node) -> Option<Node<Ms>> {
                 });
             el.attrs = attrs;
 
+            // todo etc
+            let svg_tags = vec![
+                "svg", "circle", "line", "rect"
```
MartinKavik: Can we somehow reuse `element_svg!`?

src/websys_bridge.rs
```diff
@@ -364,6 +364,16 @@ pub fn node_from_ws<Ms: Clone>(node: &web_sys::Node) -> Option<Node<Ms>> {
...
+            ];
+            let svg_tags: Vec<String> = svg_tags.into_iter().map(|t| t.to_string()).collect();
```
MartinKavik: We can inline it and crate only one Vec:
```rust
let svg_tags: Vec<String> = [
    "svg", "circle", "line", "rect"
].iter().map(ToString::to_string).collect();
```

MartinKavik: Update: I think we don't need to create `Vec` at all, 
```rust
let svg_tags = [
    "svg", "circle", "line", "rect"
];
```
should be enough.

David-OConnor: Had that originally, but couldn't get contains to work with it due to a double ref; might be another way.

David-OConnor: Running into issues with `contains` and double-references.

MartinKavik: I'm able to compile this:
```rust
    let svg_tags = [
        "svg", "circle", "line", "rect"
    ];
    let tag_name = "xxx".to_string();
    if svg_tags.contains(&tag_name.to_lowercase().as_str()) {
//        el.namespace = Some(Namespace::Svg);
    }
```
It should help you.

David-OConnor: That worked

David-OConnor: I don't like the idea of repeating the tags since it makes things more complicated when updating them, and is more error-prone, but I don't see an obvious way around it here.

MartinKavik:
> I don't see an obvious way around it here.

An idea - we can resolve it while we will be writing `/scripts/populate_tags.rs`? What do you think?

David-OConnor:
Sounds good. I'm happy to merge this and publish in the next version while we sort that out. Trying to figure out why the last cleanup commit I made isn't showing here.

MartinKavik:
I agree. Write a `todo` somewhere or create issue / add to existing one, please, so we don't forget about it.

PR #243, opened 2019-10-12
Add `#[must_use]` to RequestAnimationFrameHandle

### LINKED ISSUE #220: Add #[must_use] for RequestAnimationFrameHandle
If the handle is not properly stored after `seed::prelude::request_animation_frame`, the closure is never executed. Adding the `#[must_use]` annotation could add a warning to such unintentional handle drops

zimond:
fixes #220 

By adding this annotation, a warning is given when user accidentally drops the handle. Also I removed `www` in the Cargo.toml because `docs.rs` fails to load  github info when www is added.


David-OConnor:
Solid fix.

PR #245, opened 2019-10-15
Implement UpdateEl for &String and String

theduke:
Prevents cryptic errors when reference or owned string is passed.

### DIFF
--- a/src/dom_types.rs
+++ b/src/dom_types.rs
@@ -122,9 +122,21 @@ impl<Ms: Clone> UpdateEl<El<Ms>> for WillUnmount<Ms> {
 }
 
 impl<Ms: Clone> UpdateEl<El<Ms>> for &str {
-    // This, or some other mechanism seems to work for String too... note sure why.
     fn update(self, el: &mut El<Ms>) {
         el.children.push(Node::Text(Text::new(self.to_string())))
+    }
+}
+
+impl<Ms: Clone> UpdateEl<El<Ms>> for &String {
+    fn update(self, el: &mut El<Ms>) {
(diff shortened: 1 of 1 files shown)


David-OConnor:
Could you give an example of how to trigger the errors? I never understood why the `&str` implementation appears to work for `String` as well, and didn't add `String` because it seemed to be working as-is.

theduke:
It only seems to work in a macro context. Somehow there seems to be a dereference inserted to it always falls back to `&str`. Not sure why though.

This actually also can force redundant allocations since a String will always be cloned. (as the failing test is showing )

I stumbled on this while implementing a `jsx` proc macro for seed.

MartinKavik:
> It only seems to work in a macro context.

It works also outside macro, e.g.:
```rust
let mut el = El::empty(Tag::Div);
"hello".to_string().update(&mut el);
```

So could you provide specific example, please?

> This actually also can force redundant allocations since a String will always be cloned. (as the failing test is showing )

- Failing test?
- Optimizations and more `UpdateEl` `impl`s are in todo list - we can do it once `specialization` is stable (https://github.com/David-OConnor/seed/issues/128#issuecomment-500077596, https://github.com/David-OConnor/seed/pull/178#issue-295539411)

David-OConnor:
I can't see any harm this addition would do, and it sounds like it could help avoid problems in cases I haven't considered... I'd just like more info on what those cases are.

MartinKavik:
@theduke is it still a problem for you or we can close it?

David-OConnor:
@theduke  we'll reopen if you want to re-attack this.

PR #252, opened 2019-10-17
Added hashchange listener

David-OConnor:
Need to make `Url` parsing from string work with hash and search before merging; currently treats them as part of `path`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## v0.4.2
 - Added an `Init` struct, which can help with initial routing (Breaking)
+- The `routes` function now returns an `Option<Msg>` (Breaking)
 - updated `Tag::from()` to accept more input types
 - Improved error-handling
 - Macro `custom!` checks if you set tag, and panics when you forget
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -2,9 +2,14 @@
 
 ## v0.4.2
 - Added an `Init` struct, which can help with initial routing (Breaking)
+- Changed `Tag::Custom` to `Tag::from` (Breaking)
```
MartinKavik: I've updated `from` function to accept more input types, no other changes, i.e. no breaking changes.

David-OConnor:
Should now work with hash and search, but is still double-firing; I suspect the `hashchange` event's triggering a `popstate`.

MartinKavik:
@David-OConnor Ping me before merge, please, I would like to do review.

David-OConnor:
Ready now, assuming it works on @flosse's end.

flosse:
> Ready now, assuming it works on @flosse's end.

@David-OConnor thanks a lot, now it works!

MartinKavik:
I think we can merge it so we have this fix  in `master`. We don't have wait for other PRs because of https://github.com/David-OConnor/seed/pull/253#issuecomment-544924580.

PR #257, opened 2019-10-22
Add `MinLength` attribute.

### LINKED ISSUE #256: Attribute `minlength` missing.
The attribute `minlength` is missing.
Used for `<input type="text">` and such.

See corresponding documentation [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-minlength).

daxpedda:
See title.
Fixes #256.


David-OConnor:
Nice.

PR #259, opened 2019-10-24
Remove the Clone bound from messages where possible.

AlterionX:
There was a change that could eliminate most of the times `Msg` needed to have the `Clone` bound.

### DIFF
--- a/src/dom_types.rs
+++ b/src/dom_types.rs
@@ -67,100 +67,100 @@ pub trait UpdateEl<T> {
     fn update(self, el: &mut T);
 }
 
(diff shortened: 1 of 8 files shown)


David-OConnor:
Need to test, but this would be great from an API perspective. As you've discovered, the mandatory clone's a byproduct of implementation - not something desired!

AlterionX:
There's a `derive(Clone)` that was enforcing the trait bound on the message, so this removes that, if that's what you meant.

There's still one place where the message has the `Clone` bound that I wasn't able to remove easily, but the one that I couldn't remove has a trait bound that's pretty involved and this one hits a ton of files already, so I wanted to work on that one separately.

PR #260, opened 2019-10-25
Add test case to prove attribue 'disabled' bug

flosse:


MartinKavik:
I suggest to close this PR. We can use the code later in a new PR, when we will be updating `attrs!` implementation.

flosse:
>I suggest to close this PR. We can use the code later in a new PR, when we will be updating `attrs!` implementation.

That‘s ok for me.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #267, opened 2019-10-29
Add convenience fn to access web_sys::HtmlDocument

flosse:


src/util.rs
```diff
@@ -34,6 +34,16 @@ pub fn document() -> web_sys::Document {
         .expect("Can't find the window's document")
 }
 
+/// Convenience function to access the `web_sys::HtmlDocument`.
+pub fn html_document() -> web_sys::HtmlDocument {
+    to_html_document(document())
+}
+
+/// Convenience function to convert a `Document` to an `HtmlDocument`.
+pub fn to_html_document(doc: web_sys::Document) -> web_sys::HtmlDocument {
```
MartinKavik: - What's your use-case, please? (Can it be private?)
- Implementing `From<Document> for HtmlDocument` is probably not possible, but can we at least rename this function to something more expressive (like `document_to_html_document`)? Or more elegant solution could be to write trait like `ToHtmlDocument` with one function `to_html_document`, implement it for `Document` and add it into Seed's prelude.

src/util.rs
```diff
@@ -34,6 +34,16 @@ pub fn document() -> web_sys::Document {
         .expect("Can't find the window's document")
 }
 
+/// Convenience function to access the `web_sys::HtmlDocument`.
+pub fn html_document() -> web_sys::HtmlDocument {
```
MartinKavik: Could you provide an example? (I'm just curious what forces you to use `HtmlDocument` instead of `Document`).

flosse: > Could you provide an example?

sure: `html_document().cookie()`

> (I'm just curious what forces you to use `HtmlDocument` instead of `Document`).

the [`cookie` method](https://docs.rs/web-sys/0.3.29/web_sys/struct.HtmlDocument.html#method.cookie) is only available in `HtmlDocument`.

src/util.rs
flosse: >* What's your use-case, please? 

I need to check the cookies of the document
> (Can it be private?)

yes, this one can be private

>* Implementing `From<Document> for HtmlDocument` is probably not possible, but can we at least rename this function to something more expressive (like `document_to_html_document`)?

yes, that's possible :)

> Or more elegant solution could be to write trait like `ToHtmlDocument` with one function `to_html_document`, implement it for `Document` and add it into Seed's prelude.

that's also nice.
What do you prefer?

flosse: > I prefer `From<Document> for HtmlDocument` of course I'd also prefer this but that's not allowed in Rust. We have either to wrap one of both into a new struct or define or create feature request for `web_sys`

MartinKavik: > of course I'd also prefer this but that's not allowed in Rust.

I thought so but then you wrote "yes, that's possible :)"..  a little misunderstanding :)

I have no preference between renaming and adding trait - the first is simple and explicit, the second one doesn't pollute public API. Or choose another solution if you find it more suitable.

David-OConnor:
Leaning towards this being good: The existing methods are included because they're used internally. But they're also documented as convenient shortcuts. These are `web_sys` items with no direct link to `seed`. But it doesn't hurt anything, and prevents users from having to include this messy-code themselves, which is consistent with the project's overall philosophy.

src/util.rs
```diff
@@ -44,6 +44,16 @@ pub fn history() -> web_sys::History {
     window().history().expect("Can't find history")
 }
 
+/// Convenience function to access the `web_sys::HtmlDocument`.
+pub fn html_document() -> web_sys::HtmlDocument {
+    document_to_html_document(document())
+}
+
+/// Convenience function to convert a `Document` to an `HtmlDocument`.
+fn document_to_html_document(doc: web_sys::Document) -> web_sys::HtmlDocument {
```
MartinKavik: I would inline `document_to_html_document` body to `html_document` because `document_to_html_document` is private one-liner and called in one place.. but yeah, it's not very important :)

flosse: done :)

PR #269, opened 2019-10-29
Move `AppBuilder` and a few type aliases into a separate file.

AlterionX:
A part of the larger PR #235.


David-OConnor:
Nice split; `vdom` has always been a module that that didn't have well-defined edges of what should be in it.

PR #270, opened 2019-10-29
Extract conversion of `web_sys::El` into `El` from a method

AlterionX:
A part of #235.

This is so that the conversion can be used without the other two branches and so that it is not wrapped in a Node.


David-OConnor:
More idiomatic too.

PR #272, opened 2019-10-30
Cookies

flosse:


src/util.rs
```diff
@@ -44,6 +44,25 @@ pub fn history() -> web_sys::History {
...
+    wasm_bindgen::JsValue::from(document()).unchecked_into::<web_sys::HtmlDocument>()
+}
+
+/// Convenience function to get all cookies from the current `HtmlDocument`
+#[cfg(feature = "cookies")]
+pub fn cookies() -> cookie::CookieJar {
+    let mut jar = cookie::CookieJar::new();
+    if let Ok(cookie_str) = html_document().cookie() {
+        for cs in cookie_str.split(";") {
+            if let Ok(c) = cookie::Cookie::parse_encoded(cs.to_string()) {
```
MartinKavik: Is `cs.to_string()` necessary? `parse_encoded` accepts `Into<Cow<'c, str>>` so maybe `cs.into()` is enough and we don't have to allocate memory for `String`.

src/util.rs
```diff
@@ -44,6 +44,25 @@ pub fn history() -> web_sys::History {
...
 }
 
+/// Convenience function to access the `web_sys::HtmlDocument`.
+pub fn html_document() -> web_sys::HtmlDocument {
+    wasm_bindgen::JsValue::from(document()).unchecked_into::<web_sys::HtmlDocument>()
+}
+
+/// Convenience function to get all cookies from the current `HtmlDocument`
+#[cfg(feature = "cookies")]
+pub fn cookies() -> cookie::CookieJar {
```
MartinKavik: Can `html_document().cookie()` and `::parse_encoded(..)` fail in a real-world scenario? => Is it ok that we ignore errors?

src/util.rs
```diff
@@ -44,6 +44,25 @@ pub fn history() -> web_sys::History {
     window().history().expect("Can't find history")
 }
 
+/// Convenience function to access the `web_sys::HtmlDocument`.
+pub fn html_document() -> web_sys::HtmlDocument {
+    wasm_bindgen::JsValue::from(document()).unchecked_into::<web_sys::HtmlDocument>()
+}
+
+/// Convenience function to get all cookies from the current `HtmlDocument`
+#[cfg(feature = "cookies")]
```
MartinKavik: What's the reason for feature flag? Is app too big with `cookie` dependency?

src/util.rs
flosse: > Is `cs.to_string()` necessary?

yes, otherwise I get `borrowed value does not live long enough`

src/util.rs
flosse: > What's the reason for feature flag?

I just don't want to force all users to depend on `cookie`

> Is app too big with `cookie` dependency?

no.

src/util.rs
flosse: > Can `html_document().cookie()` and `::parse_encoded(..)` fail in a real-world scenario?

I have no idea :-\

> => Is it ok that we ignore errors?

We could at least return an `Option<CookieJar>`

MartinKavik: > We could at least return an Option<CookieJar>

Yeah, users should at least have to add `uwrap` and think about it. Or alternative is `Result<CookieJar, ()> so we can add more details in the error in the future without "big" breaking change.

src/util.rs
MartinKavik: > I just don't want to force all users to depend on cookie

I don't see problem with it . Otherwise we would have to document it, it would be another obstacle for new Seed users and code would be a little bit harder to reason about. I suggest to remove that feature flag and revisit it later, when we will be resolving https://github.com/David-OConnor/seed/issues/248.
What do you think?

flosse: :+1:

src/util.rs
flosse: I changed it to `Option` for now.

The be more precise we need `Result<Option<CookieJar>, ()>` or `Result<Vec<Cookie>, ()>` because I don't want to return a `CookieJar` that is empty.

src/util.rs
```diff
@@ -44,6 +44,28 @@ pub fn history() -> web_sys::History {
...
+}
+
+/// Convenience function to get all cookies from the current `HtmlDocument`
+pub fn cookies() -> Option<cookie::CookieJar> {
+    html_document().cookie().ok().and_then(|cookie_str| {
+        let cookies = cookie_str
+            .split(";")
+            .map(|x| x.to_owned())
+            .filter_map(|cs| cookie::Cookie::parse_encoded(cs).ok());
+        if cookies.clone().count() > 0 {
```
MartinKavik: Maybe a little bit faster / readable code:
```rust
/// Convenience function to get all cookies from the current `HtmlDocument`
/// _Note:_ Returns `None` if parsing cookies fails or there are no cookies.
pub fn cookies() -> Option<cookie::CookieJar> {
    let cookies_str = html_document().cookie().ok()?;
    let mut jar = cookie::CookieJar::new();

    for cookie_str in cookies_str.split(';') {
        let cookie = cookie::Cookie::parse_encoded(cookie_str).ok()?;
        jar.add(cookie.into_owned());
    }

    let jar_is_empty = jar.iter().next().is_none();
    if jar_is_empty {
        None
    } else {
        Some(jar)
    }
}
```
Notes:
- It also returns `None` if it fails to parse cookies.
- I assume that creating a new jar instance is a trivial operation.

David-OConnor:
Added to docs under the `misc` page.

PR #274, opened 2019-10-30
Tweak bootstrap order so that `main_el_vdom` is initialized first.

AlterionX:
Part of #235.

Forces `process_cmd_and_msg_queue` to occur after the `main_el_vdom` is initialized, making it consistent with all other calls of `process_cmd_and_msg_queue`.

Also removes an extra `Node` wrapping for calling `assign_ws_nodes_to_el` when bootstrapping the vdom.

### DIFF
--- a/src/vdom.rs
+++ b/src/vdom.rs
@@ -195,7 +195,7 @@ impl<Ms, Mdl, ElC: View<Ms> + 'static, GMs: 'static> App<Ms, Mdl, ElC, GMs> {
         }
     }
 
-    fn bootstrapped_vdom(&self) -> El<Ms> {
+    fn bootstrap_vdom(&self) -> El<Ms> {
         // "new" name is for consistency with `update` function.
         // this section parent is a placeholder, so we can iterate over children
         // in a way consistent with patching code.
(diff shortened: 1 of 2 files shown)


src/vdom.rs
```diff
@@ -195,48 +195,43 @@ impl<Ms, Mdl, ElC: View<Ms> + 'static, GMs: 'static> App<Ms, Mdl, ElC, GMs> {
...
-    pub fn run(self) -> Self {
-        self.process_cmd_and_msg_queue(
-            self.cfg
-                .initial_orders
-                .replace(None)
-                .expect("initial_orders should be set in AppBuilder::finish")
-                .effects,
-        );
-        // Our initial render. Can't initialize in new due to mailbox() requiring self.
+    fn bootstrapped_vdom(&self) -> El<Ms> {
```
MartinKavik: Can we rename it to `bootstrap_vdom`? I would assume that `bootstrapped_vdom` is only a getter.

AlterionX: Sure.

David-OConnor:
@AlterionX : Does this wrap up this stage of PRs? If not, of you think of something later, we can always add more.

AlterionX:
There's one more piece. Almost done.

PR #275, opened 2019-11-01
Added `AppBuilder.start()`, combining `finish()` and `run()`

David-OConnor:
Added alternative starting code: `AppBuilder.start()`, which combines `.finish()` and `.run()`. These two are usually associated with each other. This flow's a little cleaner.

```rust
#[wasm_bindgen(start)]
pub fn render() {
    seed::App::build(|_, _| Init::new(Model::default()), update, view)
        .finish()
        .run();
}
```

 to 

```rust
#[wasm_bindgen(start)]
pub fn render() {
    seed::App::build(|_, _| Init::new(Model::default()), update, view)
        .start();
}
```

Are there any cases where we do something between `finish()` and `run()`? Can still do the former, but am wondering for documentation purposes.

Open to alternatives / further simplifications / renames etc. For example, what do you think about `seed::build` to start, instead of `seed::App::build`? This would be a little simpler, but hides that it's instantiating an `App`.

Overall intent is to minimize boilerplate, if we can do it with minimal drawbacks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ Added a `hashchange` listener
 - Added a routing listener for changed hash
 - Fixed a namespace bug with adding children to `Svg` elements
 - Fixed a bug affecting Safari
-- Added a `seed::html_document()` and `seed::cookies` convenience functions
+- Added `seed::html_document()` and `seed::cookies` convenience functions
 
 ## v0.4.1
 - Added more SVG `At` variants
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,11 @@
 - Added an `Init` struct, which can help with initial routing (Breaking)
 - The `routes` function now returns an `Option<Msg>` (Breaking)
 - Updated `Tag::from()` to accept more input types
-- `style!` now accepts types that implement the new `ToCssValue` trait
+- `style!` now accepts also `Option<impl ToString>`
 - Fixed a bug affecting element render order
-Added a `hashchange` listener
+- Added a `hashchange` listener
 - Improved error-handling
-- Tweakd bootstrap order so that `main_el_vdom` is initialized first (internal)
+- Tweaked bootstrap order so that `main_el_vdom` is initialized first (internal)
 - Macro `custom!` checks if you set tag, and panics when you forget
 - Fixed a bug with children being absent from cloned elements
 - Improved debugging
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,16 +3,19 @@
 ## v0.4.2
 - Added an `Init` struct, which can help with initial routing (Breaking)
 - The `routes` function now returns an `Option<Msg>` (Breaking)
-- updated `Tag::from()` to accept more input types
+- Updated `Tag::from()` to accept more input types
+- `style!` now accepts types that implement the new `ToCssValue` trait
 - Fixed a bug affecting element render order
 Added a `hashchange` listener
 - Improved error-handling
+- Tweakd bootstrap order so that `main_el_vdom` is initialized first (internal)
 - Macro `custom!` checks if you set tag, and panics when you forget
 - Fixed a bug with children being absent from cloned elements
 - Improved debugging
 - Added a routing listener for changed hash
 - Fixed a namespace bug with adding children to `Svg` elements
 - Fixed a bug affecting Safari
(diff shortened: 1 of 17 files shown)


MartinKavik:
> which combines .finish() and .run()

https://github.com/David-OConnor/seed/issues/250#issuecomment-544512869
=> `start` vs `build_and_run` - I chose `build_and_run` because `builder()...build()` is a common pattern in Rust (see the end of https://github.com/David-OConnor/seed/issues/250#issuecomment-543983556) so IDE's autocomplete can help experienced Rust programmers who know that pattern and it's a little bit more expressive imho. And when we need to split `build` and `run` phase, it would be easier + it should be a little bit less painful breaking change for current users.

> seed::build to start, instead of seed::App::build

As you said - it hides that it's `App`. And it pollutes "global" `seed::` namespace. So you are losing context and there can be problems in the future once we decide to write another important builder for some other entities.

David-OConnor:
I agree with you on not using `seed::build`, and leaving `App` exposed.

What about #[wasm_bindgen(start)]
pub fn render() {
    seed::App::new(|_, _| Init::new(Model::default()), update, view)
        .start();
}?

I'm looking at the builder pattern as an implementation detail that allows some details (`routes`, `before_mount` etc) to be optional, and `start` (or something else) is simple, vice a formal pattern.

src/vdom/builder.rs
```diff
@@ -183,4 +180,10 @@ impl<Ms, Mdl, ElC: View<Ms> + 'static, GMs: 'static> Builder<Ms, Mdl, ElC, GMs>
 
         app
     }
+
+    /// Start the app; wraps `.finish()` and `.start()`.
```
MartinKavik: I think that it should be:
```rust
/// Start the app; wraps `.finish()` and `.run()`.
```
But I suggest to change it to:
```rust
/// Build and run the app.
```
and mark `finish` and `run` as deprecated.

---

I was googling a little bit to make sure that we don't make mistake with `build_and_run` (vs e.g. `start`) - and I can confirm that `build` is the most used final method for builders and it's used also in https://github.com/colin-kiegel/rust-derive-builder, so I still think that it's a good idea.
`build_and_run` vs `build_and_start` - `start` sounds better for application but our old API has method `run` - I don't have preference here.

CHANGELOG.md
```diff
@@ -3,16 +3,19 @@
 ## v0.4.2
 - Added an `Init` struct, which can help with initial routing (Breaking)
 - The `routes` function now returns an `Option<Msg>` (Breaking)
-- updated `Tag::from()` to accept more input types
+- Updated `Tag::from()` to accept more input types
+- `style!` now accepts types that implement the new `ToCssValue` trait
 - Fixed a bug affecting element render order
 Added a `hashchange` listener
 - Improved error-handling
+- Tweakd bootstrap order so that `main_el_vdom` is initialized first (internal)
```
MartinKavik: Tweaked

CHANGELOG.md
```diff
@@ -3,16 +3,19 @@
 ## v0.4.2
 - Added an `Init` struct, which can help with initial routing (Breaking)
 - The `routes` function now returns an `Option<Msg>` (Breaking)
-- updated `Tag::from()` to accept more input types
+- Updated `Tag::from()` to accept more input types
+- `style!` now accepts types that implement the new `ToCssValue` trait
 - Fixed a bug affecting element render order
 Added a `hashchange` listener
```
MartinKavik: -

CHANGELOG.md
```diff
@@ -3,16 +3,19 @@
 ## v0.4.2
 - Added an `Init` struct, which can help with initial routing (Breaking)
 - The `routes` function now returns an `Option<Msg>` (Breaking)
-- updated `Tag::from()` to accept more input types
+- Updated `Tag::from()` to accept more input types
+- `style!` now accepts types that implement the new `ToCssValue` trait
```
MartinKavik: `ToCssValue` has been renamed to `ToCSSValueForCSSValue` and documentation is hidden.
So I suggest to change the line to something like 
```
- `style!` now accepts also `Option<impl ToString>`
```
---
I don't know if we want to add public trait `ToCSSValue` so users can implement it for custom types. If we want to discuss it I suggest to create a new issue.

src/vdom/builder.rs
David-OConnor: `build_and_start` works.

PR #283, opened 2019-11-06
Added canvas helper and example

David-OConnor:
Added a canvas example, and helper functions to reduce boilerplate. 

Failing clippy due to an unused Msg; probably needs a clippy ignore tag.


src/util.rs
```diff
@@ -48,6 +48,27 @@ pub fn history() -> web_sys::History {
...
     wasm_bindgen::JsValue::from(document()).unchecked_into::<web_sys::HtmlDocument>()
 }
+/// Convenience function to access the `web_sys::HtmlCanvasElement`.
+pub fn canvas(id: &str) -> web_sys::HtmlCanvasElement {
+    let canvas = document().get_element_by_id(id).unwrap();
+    canvas
+        .dyn_into::<web_sys::HtmlCanvasElement>()
+        .map_err(|_| ())
+        .unwrap()
+}
```
MartinKavik: I'd rather return `Option<HtmlCanvasElement>` or `Result` and write comment when the returned value is `None`. 
Example:
https://github.com/David-OConnor/seed/blob/f21c80c460349823e5658cc8655f532f4771830e/src/util.rs#L54
Unwraps / expects make sense for document or body, but I don't trust the user with the correct `id`.

examples/canvas/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+fn update(_msg: Msg, _model: &mut Model, _: &mut impl Orders<Msg>) {}
+
+// View
+
+fn view(_model: &Model) -> impl View<Msg> {
+    vec![
+        h1!["Example canvas"],
+        canvas![
+            attrs![
+                At::Id => "canvas",
```
MartinKavik: I think we should replace "canvas" here in `view` and in function `draw` with `const CANVAS_ID..` to teach users the best practices.

examples/canvas/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+                At::Width => px(200),
+                At::Height => px(100),
+            ],
+            style![
+                St::Border => "1px solid #000000",
```
MartinKavik: `#000000` vs `black`

src/util.rs
```diff
@@ -48,6 +48,27 @@ pub fn history() -> web_sys::History {
...
+    canvas: &web_sys::HtmlCanvasElement,
+    context: &str,
+) -> web_sys::CanvasRenderingContext2d {
+    canvas
+        .get_context(context)
+        .unwrap()
+        .unwrap()
+        .dyn_into::<web_sys::CanvasRenderingContext2d>()
+        .unwrap()
+}
```
MartinKavik: I suggest to rename this function to `canvas_context_2d` and remove parameter `context`. Because user can send only `"2d"` as the `context` otherwise it fails during casting (https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/getContext).

I assume that the updated function shouldn't fail, but I recommend to at least replace `unwrap` with `expect` (if it make sense for you).

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## v0.4.3 [unreleased]
```
MartinKavik: Isn't `## [unreleased]` enough? How do you know that the next version will be `0.4.3` and not `0.5.0`?

MartinKavik:
> Failing clippy due to an unused Msg; probably needs a clippy ignore tag.

1. option
```rust
#[allow(clippy::empty_enum)]
#[derive(Copy, Clone)]
enum Msg {}
```
2. option
```rust
type Msg = ();
```

examples/canvas/src/lib.rs
```diff
@@ -0,0 +1,53 @@
...
+fn update(_msg: Msg, _model: &mut Model, _: &mut impl Orders<Msg>) {}
+
+// View
+
+fn view(_model: &Model) -> impl View<Msg> {
+    vec![
+        h1!["Example canvas"],
+        canvas![
+            attrs![
+                At::Id => "canvas",
```
MartinKavik: `CANVAS_ID`

examples/canvas/src/lib.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    ctx.move_to(0., 0.);
+    ctx.line_to(200., 100.);
+    ctx.stroke();
+}
+
+#[wasm_bindgen(start)]
+pub fn render() {
+    seed::App::build(|_, _| Init::new(Model {}), update, view).build_and_start();
+    draw();
```
MartinKavik: I assume that your masterpiece will vanish with the second render :) - can we add it somehow into the render loop? (I think that it will be the first question posted when somebody see this example.)

We can extend the example - e.g add button "Change color!" so we can test it.

src/util.rs
David-OConnor: Good point. Originally had it called that, but assumed this one was more general, without checking the casting.

MartinKavik:
I think we can merge it. 
However I'm still not very happy with the draw & render loop, but we can't do it properly with the current API - I'll try to refactor it once I resolve https://github.com/David-OConnor/seed/issues/207.

Issue #250, opened 2019-10-17
New AppBuilder API


MartinKavik:
@MartinKavik @David-OConnor I'm starting to get half a mind to move the `Init` from `build` to `run` or something similar to isolate all the side effects inside of `run` instead of distributing the stateful effects across `finish` and `run`.

I think that this is better since we do this weird dance around `initial_orders` right now where we don't quite have the correct state to do what we have half the state we want inside `Builder` and half the state inside the `App`.

So we would have
```rust
App::build(update, view).finish().run_with_init(|_, _| {
    // Do things with url + orders.
    // `Init::default` is a wrapper for `Init::new(Model::default())`.
    Init::default()
})
```
or
```rust
App::build(update, view).finish().run() // Calls `Init::default` by default.
```

It's a bit of a departure from what we currently have, but I think that it makes the implementation a little less weird and the usage clearer about when exactly the `InitFn` happens (since it seems to run half in `finish` and half in `run` right now).

Personally, I also think that `mount` should also be inside `Init` since we might want to mount it to a different place depending on the initial route, but we seem to keep that around forever, so I changed my mind. Point is, there are parts of `AppCfg` that are not immutable, and I would like to change that.

The downsides, I think, are also fairly obvious in that it's a breaking change, and so on. Also, the model is now provided in the `run` method instead of the `build` method, which is also... unique, so to speak.

I haven't seen if the types work out yet, but it seems reasonable to me.

_Originally posted by @AlterionX in https://github.com/David-OConnor/seed/pull/235#issuecomment-542944604_

MartinKavik:
> isolate all the side effects inside of run instead of distributing the stateful effects across `finish` and `run`

There are more hidden side-effects in `AppBuilder`'s methods - see e.g. `mount`.

> `App::build(update, view).finish().run()`

- `App::build(update, view).run()` would be even nicer.
- I'm not sure if compiler allows you to write `::default()` function, because implementation of `Default` trait for `Model` is optional.  

> It's a bit of a departure from what we currently have, but I think that it makes the implementation a little less weird and the usage clearer about when exactly the InitFn happens (since it seems to run half in finish and half in run right now).

I agree. I've added `init` as the part of the bigger PR (https://github.com/David-OConnor/seed/pull/189). `AppBuilder` code wasn't super clean before this change and I've made it worse as a trade-off for fewer breaking changes.

>  mount it to a different place depending on the initial route

Interesting. Do you have a real-world use-case?

---

[...]

PR #290, opened 2019-11-20
New AppBuilderAPI

AlterionX:
Here's an implementation for the new API (#250) that I've been toying around with.

I may have gone a little crazy on the types.... I could have also combined 4 of the fields in the new `AppBuilder` struct into 2, but it might have been (even more) confusing, so I left it as is.

It should still support everything for the older version, and I haven't marked the old methods as deprecated yet.

I also haven't gotten around to converting the examples yet, but I've been up pretty much all night at this point, so I'll fix it later.

### DIFF
--- a/examples/counter/src/lib.rs
+++ b/examples/counter/src/lib.rs
@@ -105,5 +105,5 @@ fn view(model: &Model) -> impl View<Msg> {
 
 #[wasm_bindgen(start)]
 pub fn render() {
-    seed::App::build(|_, _| Init::new(Model::default()), update, view).build_and_start();
+    seed::App::builder(update, view).build_and_start();
 }
--- a/src/orders.rs
+++ b/src/orders.rs
@@ -110,6 +110,11 @@ impl<Ms, Mdl, ElC: View<Ms>, GMs> OrdersContainer<Ms, Mdl, ElC, GMs> {
             app,
         }
     }
+
+    pub(crate) fn append_from(&mut self, mut other: Self) {
+        self.should_render = other.should_render;
+        self.effects.append(&mut other.effects);
+    }
 }
 
 impl<Ms: 'static, Mdl, ElC: View<Ms> + 'static, GMs> Orders<Ms, GMs>
--- a/src/orders.rs
+++ b/src/orders.rs
@@ -111,7 +111,7 @@ impl<Ms, Mdl, ElC: View<Ms>, GMs> OrdersContainer<Ms, Mdl, ElC, GMs> {
         }
     }
 
-    pub(crate) fn append_from(&mut self, mut other: Self) {
+    pub(crate) fn merge(&mut self, mut other: Self) {
         self.should_render = other.should_render;
         self.effects.append(&mut other.effects);
     }
--- a/src/vdom.rs
+++ b/src/vdom.rs
@@ -18,7 +18,10 @@ pub use alias::*;
 
 // Building process.
 pub mod builder;
-pub use builder::{Builder as AppBuilder, Init, InitFn, MountPoint, MountType, UrlHandling};
+pub use builder::{
+    AfterMount, BeforeMount, Builder as AppBuilder, Init, InitFn, MountPoint, MountType,
+    UrlHandling,
+};
 
 use crate::{
     dom_types::{self, El, MessageMapper, Namespace, Node, View},
@@ -171,12 +174,17 @@ impl<Ms: 'static, Mdl: 'static, ElC: View<Ms>, GMs> ::std::fmt::Debug for App<Ms
     }
 }
 
+#[deprecated(since = "0.4.3", note = "Part of the old Init API.")]
 type InitAppBuilder<Ms, Mdl, ElC, GMs> =
     AppBuilder<Ms, Mdl, ElC, GMs, builder::MountPointInitInitAPI<(), InitFn<Ms, Mdl, ElC, GMs>>>;
 
 /// We use a struct instead of series of functions, in order to avoid passing
 /// repetitive sequences of parameters.
 impl<Ms, Mdl, ElC: View<Ms> + 'static, GMs: 'static> App<Ms, Mdl, ElC, GMs> {
+    #[deprecated(
+        since = "0.4.3",
+        note = "Use `builder` with `AppBuilder::{after_mount, before_mount}` instead."
+    )]
     pub fn build(
         init: impl FnOnce(routing::Url, &mut OrdersContainer<Ms, Mdl, ElC, GMs>) -> Init<Mdl> + 'static,
         update: UpdateFn<Ms, Mdl, ElC, GMs>,
@@ -275,10 +283,6 @@ impl<Ms, Mdl, ElC: View<Ms> + 'static, GMs: 'static> App<Ms, Mdl, ElC, GMs> {
             new.children = dom_nodes.children;
         }
 
-        self.setup_window_listeners();
-        patch::setup_input_listeners(&mut new);
-        patch::attach_listeners(&mut new, &self.mailbox());
-
         // Recreate the needed nodes. Only do this if requested to takeover the mount point since
         // it should only be needed here.
         if mount_type == MountType::Takeover {
@@ -325,10 +329,9 @@ impl<Ms, Mdl, ElC: View<Ms> + 'static, GMs: 'static> App<Ms, Mdl, ElC, GMs> {
             mount_type,
             into_after_mount,
             ..
-        } = self
-            .run_cfg
-            .take()
-            .expect("run_cfg should be set in App::new which is called from AppBuilder");
+        } = self.run_cfg.take().expect(
+            "run_cfg should be set in App::new which is called from AppBuilder::build_and_start",
+        );
 
         // Bootstrap the virtual DOM.
         self.data
@@ -362,6 +365,13 @@ impl<Ms, Mdl, ElC: View<Ms> + 'static, GMs: 'static> App<Ms, Mdl, ElC, GMs> {
             }
             UrlHandling::None => (),
         };
+
+        self.setup_window_listeners();
+        patch::setup_input_listeners(&mut self.data.main_el_vdom.borrow_mut().as_mut().unwrap());
+        patch::attach_listeners(
+            self.data.main_el_vdom.borrow_mut().as_mut().unwrap(),
+            &self.mailbox(),
+        );
 
         // Update the state on page load, based
         // on the starting URL. Must be set up on the server as well.
--- a/src/vdom.rs
+++ b/src/vdom.rs
@@ -349,14 +349,14 @@ impl<Ms, Mdl, ElC: View<Ms> + 'static, GMs: 'static> App<Ms, Mdl, ElC, GMs> {
         match url_handling {
             UrlHandling::PassToRoutes => {
                 let url = routing::current_url();
-                if let Some(routes) = self.data.routes.borrow().as_ref() {
-                    if let Some(routing_msg) = routes(url) {
-                        (self.cfg.update)(
-                            routing_msg,
-                            &mut self.data.model.borrow_mut().as_mut().unwrap(),
-                            &mut orders,
-                        );
-                    }
+                let routing_msg = self
+                    .data
+                    .routes
+                    .borrow()
+                    .as_ref()
+                    .and_then(|routes| routes(url));
+                if let Some(routing_msg) = routing_msg {
+                    orders.effects.push_back(routing_msg.into());
                 }
             }
             UrlHandling::None => (),
--- a/src/vdom.rs
+++ b/src/vdom.rs
@@ -125,10 +125,10 @@ pub struct AppData<Ms: 'static, Mdl> {
     pub render_timestamp: Cell<Option<RenderTimestamp>>,
 }
 
-type OptDynRunCfg<Ms, Mdl, ElC, GMs> =
-    Option<AppRunCfg<Ms, Mdl, ElC, GMs, dyn builder::IntoAfterMount<Ms, Mdl, ElC, GMs>>>;
+type OptDynInitCfg<Ms, Mdl, ElC, GMs> =
+    Option<AppInitCfg<Ms, Mdl, ElC, GMs, dyn builder::IntoAfterMount<Ms, Mdl, ElC, GMs>>>;
 
-pub struct AppRunCfg<Ms, Mdl, ElC, GMs, IAM: ?Sized>
+pub struct AppInitCfg<Ms, Mdl, ElC, GMs, IAM: ?Sized>
 where
     Ms: 'static,
     Mdl: 'static,
@@ -161,7 +161,7 @@ where
     ElC: View<Ms>,
 {
     /// Temporary app configuration that is removed after app begins running.
-    pub run_cfg: OptDynRunCfg<Ms, Mdl, ElC, GMs>,
+    pub init_cfg: OptDynInitCfg<Ms, Mdl, ElC, GMs>,
     /// App configuration available for the entire application lifetime.
     pub cfg: Rc<AppCfg<Ms, Mdl, ElC, GMs>>,
     /// Mutable app state
@@ -216,13 +216,13 @@ impl<Ms, Mdl, ElC: View<Ms> + 'static, GMs: 'static> App<Ms, Mdl, ElC, GMs> {
         mount_point: Element,
         routes: Option<RoutesFn<Ms>>,
         window_events: Option<WindowEvents<Ms, Mdl>>,
-        run_cfg: OptDynRunCfg<Ms, Mdl, ElC, GMs>,
+        init_cfg: OptDynInitCfg<Ms, Mdl, ElC, GMs>,
     ) -> Self {
         let window = util::window();
         let document = window.document().expect("Can't find the window's document");
 
         Self {
-            run_cfg,
+            init_cfg,
             cfg: Rc::new(AppCfg {
                 document,
                 mount_point,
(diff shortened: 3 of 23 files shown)


AlterionX:
```rust
    seed::App::builder(update, view).build_and_run();
```
This is what counter looks like now. Also, I think I need to combine `build_and_run` and `build_and_start`.

David-OConnor:
Love that - Adding the `init` system was necessary to make things more flexible, but the existing API had too much boilerplate for default settings.

examples/counter/src/lib.rs
```diff
@@ -105,5 +105,5 @@ fn view(model: &Model) -> impl View<Msg> {
 
 #[wasm_bindgen(start)]
 pub fn render() {
-    seed::App::build(|_, _| Init::new(Model::default()), update, view).build_and_start();
+    seed::App::builder(update, view).build_and_run();
```
MartinKavik: Why you changed `build_and_start` to `build_and_run`?

src/vdom.rs
```diff
@@ -153,18 +170,42 @@ impl<Ms: 'static, Mdl: 'static, ElC: View<Ms>, GMs> ::std::fmt::Debug for App<Ms
...
+type InitAppBuilder<Ms, Mdl, ElC, GMs> = AppBuilder<
+    Ms,
+    Mdl,
+    ElC,
+    GMs,
+    (),
+    (),
+    (),
+    InitFn<Ms, Mdl, ElC, GMs>,
+    builder::MountPointInitInitAPI<(), InitFn<Ms, Mdl, ElC, GMs>>,
```
MartinKavik: Typo? (`MountPointInitInitAPI` vs `MountPointInitAPI`)

src/vdom.rs
```diff
@@ -119,6 +121,21 @@ pub struct AppData<Ms: 'static, Mdl> {
...
+pub struct AppRunCfg<Ms, Mdl, ElC, GMs, IAM: ?Sized>
+where
+    Ms: 'static,
+    Mdl: 'static,
+    ElC: View<Ms>,
+    IAM: builder::IntoAfterMount<Ms, Mdl, ElC, GMs>,
+{
+    mount_type: MountType,
+    into_after_mount: Box<IAM>,
+    phantom: std::marker::PhantomData<(Ms, Mdl, ElC, GMs)>,
```
MartinKavik: I suggest to add `use std::marker::PhantomData;` because `PhantomData` is a standard Rust type and to make it consistent with `builder.rs`.

examples/websocket/Cargo.toml
```diff
@@ -11,6 +11,7 @@ path = "src/client.rs"
 [[bin]]
 name = "server"
 path = "src/server.rs"
+required-features = ["server"]
```
MartinKavik: Why?

src/vdom.rs
```diff
@@ -288,9 +328,46 @@ impl<Ms, Mdl, ElC: View<Ms> + 'static, GMs: 'static> App<Ms, Mdl, ElC, GMs> {
...
+        } = into_after_mount.into_after_mount(routing::initial_url(), &mut orders);
+
+        self.data.model.replace(Some(model));
+
+        // TODO: Does this go before or after setting up the routes listener?
+        match url_handling {
+            UrlHandling::PassToRoutes => {
+                let url = routing::initial_url();
+                if let Some(r) = self.data.routes.borrow().as_ref() {
+                    if let Some(u) = r(url) {
```
MartinKavik: I don't know if it's old copy/pasted code or a new one, but those one-letter variables are hard to read.

src/vdom.rs
MartinKavik: Can we reorder types and set default values for some of them to `()`?
The goal is to have something like:
```rust
type InitAppBuilder<Ms, Mdl, ElC, GMs> = AppBuilder<
    Ms,
    Mdl,
    ElC,
    GMs,
    InitFn<Ms, Mdl, ElC, GMs>,
    builder::MountPointInitInitAPI<(), InitFn<Ms, Mdl, ElC, GMs>>,
```

src/vdom/builder.rs
```diff
@@ -1,128 +1,218 @@
...
-            panic!(
-                "Can't find element with id={:?} - app cannot be mounted!\n\
-                 (Id defaults to \"app\", or can be set with the .mount() method)",
-                self
-            )
-        })
-    }
+pub struct MountPointInitInitAPI<MP, II> {
+    mp: PhantomData<MP>,
+    ii: PhantomData<II>,
```
MartinKavik: Can use more expressive names? I have no idea what `mp` and `ii` are without context.

MartinKavik:
Looks good on the first glance. Let me know when you have questions or when it's ready for the second review.

examples/counter/src/lib.rs
AlterionX: I need to combine the two, actually. They took two different code paths that I didn't realize could be combined until right before I went to sleep.

src/vdom.rs
AlterionX: It's the `(MountPoint[And]Init)InitAPI`, so not a typo. But the name is a bit weird. I was going to name it `LegacyInitAPI` but that didn't tell me what it did, so I went with this one instead. If you have a better idea, I'd be glad to hear it, though.

examples/websocket/Cargo.toml
AlterionX: Ah, I was having trouble compiling (accidentally used `cargo make check`) so I was trying to figure things out. I must've committed this by accident.

src/vdom.rs
AlterionX: Noted. They were copy and pasted. I'll fix them soon.

src/vdom/builder.rs
AlterionX: Hrm, yeah, true. I might end up moving the 4 fields into their respective `InitAPI` structs, so this might change a bit today.

src/vdom.rs
MartinKavik: I didn't dive into the logic, so I don't know the purpose. But you are saying it can be named `LegacyInitAPI`, so I assume it will be deleted / refactored in the future. If it's true, just mark it `deprecated` or add comment like `will be removed once the new API is implemented` and we don't have to waste time on it now.

AlterionX: Managed to reduce the number of types parameters in another way.

AlterionX:
@MartinKavik I think it's ready for another review!

src/vdom.rs
```diff
@@ -288,9 +324,54 @@ impl<Ms, Mdl, ElC: View<Ms> + 'static, GMs: 'static> App<Ms, Mdl, ElC, GMs> {
...
+                        (self.cfg.update)(
+                            routing_msg,
+                            &mut self.data.model.borrow_mut().as_mut().unwrap(),
+                            &mut orders,
+                        );
+                    }
+                }
+            }
+            UrlHandling::None => (),
+        };
```
MartinKavik: I think we can move it under the comments lines:
```
// Update the state on page load, based
// on the starting URL. Must be set up on the server as well.
```
What do you think?

src/vdom.rs
```diff
@@ -313,13 +394,7 @@ impl<Ms, Mdl, ElC: View<Ms> + 'static, GMs: 'static> App<Ms, Mdl, ElC, GMs> {
...
-        self.process_cmd_and_msg_queue(
-            self.cfg
-                .initial_orders
-                .replace(None)
-                .expect("initial_orders should be set in AppBuilder::finish")
-                .effects,
-        );
+        self.process_cmd_and_msg_queue(orders.effects);
         // TODO: In the future, only run the following line if the above statement does not
         // TODO: call `rerender_vdom` for efficiency.
```
MartinKavik:
```rust
// TODO: In the future, only run the following line if the above statement:
//  - didn't force-rerender vdom
//  - didn't schedule render
//  - doesn't want to skip render 
```

src/vdom/builder/before_mount.rs
```diff
@@ -0,0 +1,121 @@
...
+pub trait MountPoint {
+    fn element(self) -> Element;
+}
+
+impl MountPoint for &str {
+    fn element(self) -> Element {
+        util::document().get_element_by_id(self).unwrap_or_else(|| {
+            panic!(
+                "Can't find element with id={:?} - app cannot be mounted!\n\
+                 (Id defaults to \"app\", or can be set with the .mount() method)",
```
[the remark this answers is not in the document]
AlterionX: The Debug impl for String should insert quotes. Display doesn't, however.

src/vdom.rs
```diff
@@ -288,9 +324,54 @@ impl<Ms, Mdl, ElC: View<Ms> + 'static, GMs: 'static> App<Ms, Mdl, ElC, GMs> {
...
         note = "Please use `AppBuilder.build_and_start` instead"
     )]
-    pub fn run(self) -> Self {
+    pub fn run(mut self) -> Self {
+        let AppRunCfg {
+            mount_type,
+            into_after_mount,
+            ..
+        } = self.run_cfg.take().expect(
+            "run_cfg should be set in App::new which is called from AppBuilder::build_and_start",
```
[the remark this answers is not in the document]
AlterionX: Oh, yes.

src/vdom.rs
```diff
@@ -288,9 +324,54 @@ impl<Ms, Mdl, ElC: View<Ms> + 'static, GMs: 'static> App<Ms, Mdl, ElC, GMs> {
...
+        );
+
         // Bootstrap the virtual DOM.
-        self.data.main_el_vdom.replace(Some(self.bootstrap_vdom()));
+        self.data
+            .main_el_vdom
+            .replace(Some(self.bootstrap_vdom(mount_type)));
+
+        // Can this simply be `self.update`?
```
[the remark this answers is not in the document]
AlterionX: Ah, I should get rid of that. I decided that it wouldn't be reasonable.

src/vdom.rs
```diff
@@ -313,13 +394,7 @@ impl<Ms, Mdl, ElC: View<Ms> + 'static, GMs: 'static> App<Ms, Mdl, ElC, GMs> {
         }
 
         // Our initial render. Can't initialize in new due to mailbox() requiring self.
```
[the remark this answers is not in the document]
AlterionX: I think the first part is? We don't actually call view until here. The latter part might be stale though.

src/vdom/builder/after_mount.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+}
+
+pub trait Into<Ms: 'static, Mdl, ElC: View<Ms>, GMs> {
+    fn into_after_mount(
+        self: Box<Self>,
+        init_url: Url,
+        orders: &mut OrdersContainer<Ms, Mdl, ElC, GMs>,
+    ) -> AfterMount<Mdl>;
+}
```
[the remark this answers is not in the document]
AlterionX: The conversion function has a few additional parameters, so I can't directly use `From` or `Into`. I can do something like `Into<(T, Url, &mut OrdersContainer<Ms, Mdl, ElC, GMs>)>`, but then it would be difficult to simply use one of the parameters (`T`) in the types, which is needed elsewhere. I'm not entirely sure how to express `allow this type, T, in the type parameters when used in conjunction with `some_long_list` of other types outside of creating a new trait.

I actually wanted to use `IntoAfterMount`, but clippy complained at me, so I ended up just deleting the trailing words instead of stopping clippy from complaining and re-exported it at a different level (`builder.rs`).

src/vdom.rs
AlterionX: If we match old behavior, this needs to happen before the window listeners are set up. If that's not necessary behavior to keep (or even behavior we shouldn't keep), then yes.

src/vdom.rs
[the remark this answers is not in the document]
AlterionX: Hrm, I stand corrected. This was always conditionally true since we call `update` if we pass `PassToRoutes` for `UrlHandling`. So it was out of date since the beginning, I guess.

src/vdom/builder/after_mount.rs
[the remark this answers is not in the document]
MartinKavik: > I actually wanted to use IntoAfterMount, but clippy complained at me

We both want to name it `IntoAfterMount` so I suggest to rename it and to allow clippy warning for this trait (and for other similar traits).

src/vdom/builder/before_mount.rs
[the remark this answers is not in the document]
MartinKavik: Yeah, sorry, I'm used to write this comment automatically in code reviews because other languages don't have this feature  :)

src/vdom.rs
MartinKavik: Don't look at the old code / behavior very much, there is a chance it was wrong.

Write it how it makes sense to you and it would be nice to add comment why you've chosen that position/order so we don't need to think about it in the future.

And it would be also nice to test it with an example, but the only one with routing is TodoMVC and it looks buggy / incomplete according to its code. 
So we should write some more examples with routing and fix TodoMVC in the future.
 (cc @David-OConnor )

MartinKavik:
@David-OConnor I will be several days without PC - so once the last comments are resolved and you don't want to add more, feel free to merge it.

src/vdom.rs
AlterionX: I think I'm going to change this. Right now, I think this function is fairly confusing. What we have here is as follows:

`IntoAfterMount` generates a series or `Effect`s.
`model` is set.
`routing` function is called and creates a message.
`update` is called on the routing message. (This message is created after the effects in init!)
Previously generated effects are called. (These messages are created before the messages in `IntoAfterMount`!)

I propose that we simply toss the routing message onto the queue and process them in the order that they are created.

David-OConnor:
@AlterionX : Do you plan to modify this further? Ready to merge. Overall a solid improvement, both with the `after_mount` functionality, and the cleaner/reduced-boilerplate API.

AlterionX:
I have a small question about the order in which messages are processed. It's somewhere in the review, but I'll post it again here:

> I think I'm going to change this. Right now, I think the order in which effects are processed is fairly confusing. What we have here is as follows:
>
> 1. `IntoAfterMount` generates a series of `Effect`s, `message_q_0`.
> 2. `model` is set.
> 3. `routing` function is called and creates a message, `message_q_1`.
> 4. `message_q_1` is processed with update. (This is created second!)
> 5. Previously generated effects (`message_q_0`) are called. (This is created first!)
>
> I propose that we simply toss the routing message onto the queue and process them in the order that they are created. This might break things. But I think that this is the order they should be processed in. If a message needs to get passed _after_ routing, then I think that the user should implement it so that the routing message triggers the additional effects.

David-OConnor:
What do you think it'll break? Your proposal sounds good.

AlterionX:
Just a bit afraid to change behavior, really. I just realized that I forgot to push the changes that actually do what I said, but I'd already implemented it. Whoops! Sorry about that.

David-OConnor:
Let's not worry about that - if we break things, that's fine. Hopefully will catch anything between the merge and the next release, and if not, no big deal.

src/orders.rs
```diff
@@ -110,6 +110,11 @@ impl<Ms, Mdl, ElC: View<Ms>, GMs> OrdersContainer<Ms, Mdl, ElC, GMs> {
             app,
         }
     }
+
+    pub(crate) fn append_from(&mut self, mut other: Self) {
+        self.should_render = other.should_render;
+        self.effects.append(&mut other.effects);
+    }
```
MartinKavik: Could we rename it to something like `merge`? 
`append_from` implies that `orders` is only a container like `Vector` and we assume the user knows that it contains `effects`.

`merge` examples: [Lodash](https://lodash.com/docs/4.17.15#merge) or [Elixir](https://hexdocs.pm/elixir/Map.html#merge/2)

src/vdom.rs
```diff
@@ -141,7 +160,9 @@ where
     Mdl: 'static,
     ElC: View<Ms>,
 {
-    /// Stateless app configuration
+    /// Temporary app configuration that is removed after app begins running.
+    pub run_cfg: OptDynRunCfg<Ms, Mdl, ElC, GMs>,
```
MartinKavik: Could we rename `run_cfg` to something like `initial_cfg` or `temp_cfg`? (And `OptDynRunCfg` accordingly).

My intuition tells me that `run_cfg` is the "main" configuration which is available for the app lifetime, which is wrong.

src/vdom.rs
MartinKavik: @AlterionX 
I think you have the deepest knowledge of this problem / know the best solution because you are refactoring it for a while. And it makes sense to me and I trust you with it - so I agree with your suggestions. And it should be tested with the real projects - it will be, once it's in `master`.

src/orders.rs
AlterionX: It's supposed to be internal (`pub(crate)`) but I don't mind.

MartinKavik: > It's supposed to be internal (pub(crate)) but I don't mind.

Yeah, but I think you'll thank your younger self as soon as you get back to this code :)

You'll hate me, but can you change it to only `merge` and not `merge_with`?
Suffix `*_with` implies that the function expects a merging function.
Examples: 
  - [Vec::resize_with](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.resize_with)
  - [_.mergeWith](https://lodash.com/docs/4.17.15#mergeWith)
  - [Option::get_or_insert_with](https://doc.rust-lang.org/core/option/enum.Option.html#method.get_or_insert_with)

AlterionX: Hahaha, my younger self.

Ah, yeah. Didn't realize about the `with`.

AlterionX:
@MartinKavik Made the change! Hopefully it's good to go.

MartinKavik:
@AlterionX 
> This branch cannot be rebased due to conflicts

Please rebase your branch onto updated `master`, squash commits as it makes sense to you and I'll shoot it to `master`. Thanks :)

MartinKavik:
@AlterionX Good job! Let's find out how it works in the real world :)

David-OConnor:
Are we ready to merge this?

MartinKavik:
@David-OConnor Sorry I was a little bit too fast, but code seems to be ok, no conflicts and tests pass.

PR #300, opened 2019-11-29
If history state is null, return initial Url

David-OConnor:


src/routing.rs
```diff
@@ -190,6 +175,7 @@ pub fn push_route<U: Into<Url>>(url: U) -> Url {
 
     util::history()
         .push_state_with_url(&data, title, Some(&path))
+//        .push_state_with_url(&url.to_str(), title, Some(&path))
```
MartinKavik: Forgotten comment?

David-OConnor: Yea, was experimenting with your suggestion from the Issue. Unrelated: I'm getting the async CI failure on my machine as well; reverting Cargo.toml for now.

src/routing.rs
```diff
@@ -214,7 +200,12 @@ pub fn setup_popstate_listener<Ms>(
             if let Some(routing_msg) = routes(url) {
                 update(routing_msg);
             }
-        };
+        } else {
+            // `ev.state()` will be a null JsValue for the landing page.
+            if let Some(routing_msg) = routes(initial_url()) {
+                update(routing_msg);
+            }
+        }
```
MartinKavik: We should use `if`s as little as possible, especially `if let` is dangerous (as we found out because of the original issue).
And I don't like to respect DRY at all cost, but in this case we can reuse block with `update(routing_msg)`.

So I suggest to change
```rust
      if let Some(state_str) = ev.state().as_string() {
            let url: Url =
                serde_json::from_str(&state_str).expect("Problem deserializing popstate state");
            // Only update when requested for an update by the user.
            if let Some(routing_msg) = routes(url) {
                update(routing_msg);
            }
        } else {
            // `ev.state()` will be a null JsValue for the landing page.
            if let Some(routing_msg) = routes(initial_url()) {
                update(routing_msg);
            }
        }
```
To something like:
```rust
        // `ev.state()` will be a null `JsValue` for the landing page.
        let url = match ev.state().as_string() {
            Some(state_str) => serde_json::from_str(&state_str).expect("Problem deserializing popstate state"),
            None => initial_url()
        };
        // Only update when requested for an update by the user.
        if let Some(routing_msg) = routes(url) {
            update(routing_msg);
        }
```

src/routing.rs
```diff
@@ -214,7 +199,12 @@ pub fn setup_popstate_listener<Ms>(
             if let Some(routing_msg) = routes(url) {
                 update(routing_msg);
             }
-        };
+        } else {
+            // `ev.state()` will be a null JsValue for the landing page.
+            if let Some(routing_msg) = routes(initial_url()) {
```
MartinKavik: Do we want to rename `initial_url` to `current_url`? I think `initial_url` doesn't make sense with this context and it gets current url under the hood.

src/routing.rs
MartinKavik: > was experimenting with your suggestion from the Issue

Don't use `data` and use current url instead?

src/routing.rs
hwchen: Interesting note about `if let` being dangerous, I've been pretty liberal with it but this issue is a good example of why exhaustive matching is good.

MartinKavik: Yeah, my colleagues thought that I'm weird when I was writing exhaustive switches also in Typescript 😃

MartinKavik:
@David-OConnor Can we merge it? We can try to use only `current_url` instead of `data` in another PR.

PR #305, opened 2019-12-03
Release v0.5.0

David-OConnor:


MartinKavik:
- Could you add to release_checklist that we should update Seed version in Readme? (As you've already done.)
- Could you add change from PR https://github.com/seed-rs/seed/pull/304 to this PR? (So we don't have to resolve conflicts.)

Issue #261, opened 2019-10-25
Initial value of disabled attribute is not rendered as expected


flosse:
```rust
button![attrs{At::Disable => false;}]
```
renders to
```html
<button disabled="false"></button>
```
instead of
```html
<button></button>
```

See #260

MartinKavik:
It's not bug but feature, see https://github.com/David-OConnor/seed/blob/afd4fa8387f40e2aaf7f07cb41a36da68f0b808e/src/dom_types/values.rs#L44

It was pain to maintain hardcoded DB of attributes which should or don't should be hidden if you set it's value to something falsy. And there were often bugs if I remember correctly.

We want to have auto-generated types for Tags, Attributes, Events and Styles. Then we want to type somehow also Attribute values and Style values - it should prevent from surprising behavior like you described. 
Styles are the only entity now, that is populated from DB to generate Rust types - DB is here: https://github.com/MartinKavik/html-css-db.

flosse:
Yea, typed values are great, but why is there still a `"false"` string when the should be nothing?

MuhannadAlrusayni:
I think the conversion should happen inside the attrs! macro some how.

MartinKavik:
@flosse:

> why is there still a "false" string when there should be nothing?

You need to call method `.as_at_value()`, e.g. `attrs!{ At::Disabled => false.as_at_value() }`. Everything is converted to `String` and taken as the valid value by default - https://github.com/David-OConnor/seed/blob/afd4fa8387f40e2aaf7f07cb41a36da68f0b808e/src/dom_types/values.rs#L59
We can't implement `ToString` for `AtValue` and `bool` for `AtValue` at the same time on the `stable` Rust, so calling `as_at_value` is a trade-off. Otherwise you would have to implement only specific types for `AtValue`.
Another reason for trait `AsAtValue` - 
https://github.com/David-OConnor/seed/blob/afd4fa8387f40e2aaf7f07cb41a36da68f0b808e/src/dom_types/values.rs#L75
\- you can implement it for your types and use - e.g. `Option<X>` instead of  `bool` - as the attribute value.

So yeah, there is space for improvements - you can try design better API or we can wait until the `specialization` is integrated into `stable` Rust.

---

Another problem with bool as attribute value - what false or true really mean? E.g.:
- attribute `disabled`
  - `false` => hide the entire attribute
  - `true` =>  `<button disabled>` or `<button disabled="disabled">` or other string?

- attribute `autocomplete`
[...]

Issue #211, opened 2019-08-31
Proposal: Typed CSS properties


MartinKavik:
Discussion in https://github.com/David-OConnor/seed/pull/209#discussion_r319706950:

> David-OConnor : 
Related: What do y'all think of having an `St` enum: An equivalent to `At`, but for styles? The groundwork's been there in a WIP state for a while; ended up not continuing due to the [large number of possible styles](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference). I don't think it would hurt anything, but haven't heard a suggestion about this. The change would be to fill out [the make_styles!](https://github.com/David-OConnor/seed/blob/master/src/dom_types.rs#L340) macro call, add `St` to the prelude, and document it.

> tiberiusferreira: 
I really like the idea, but I wonder if the increase in binary size will be negligible, after all, the text will be stored (uncompressed before gzip, I imagine) somewhere in the binary. Anyway, I'd rather have the convenience even if at the cost of some KBs.

MartinKavik:
> I wonder if the increase in binary size will be negligible

I was experimenting a little bit with [MartinKavik/seed-quickstart-webpack](https://github.com/MartinKavik/seed-quickstart-webpack) - there is a generator (PostCSS plugin) that creates Rust code from CSS classes. And because there is also TailwindCSS, the generated file `.rs` is big (86k lines).
However it seems that it doesn't affect the size of the `.wasm` file - it doesn't matter if that generated file has a struct with 5000 fields or 1, built `.wasm` always has 182 kb (~ 80kb compressed). 

tl;dr:
So I think that Rust is able to optimize it and adds only used values into the output. But we should test it once we add typed CSS into Seed for sure.

P.S.
Interesting fact - when I add `log!` into code, `wasm` file size is increased by ~200kb. We can investigate it / write into docs in the future.

MartinKavik:
> The groundwork's been there in a WIP state for a while; ended up not continuing due to the large number of possible styles.

Yeah, it's definitely work for a generator.
Source list can be https://www.w3.org/Style/CSS/all-properties.en.html (there are more formats, see the end of the page) or https://developer.mozilla.org/en-US/docs/MDN/Contribute/Howto/Update_the_CSS_JSON_DB.

I suggest to find similar source lists for other typed entities - attributes, events and tags.
So we can write generators for all at once and probably refactor it. Ideally we would run something like `cargo make update_types` to regenerate types for all entities.

---

[...]

PR #311, opened 2019-12-11
feat: Make attributes more convenient

TatriX:
This is an experimental approach to attributes.
Please take a look at test file I've added: tests/attrs.rs

### Motivation
When I was reading the guide I found `.as_at_value` which was slightly confusing.
I assume that `bool` and regular strings are more common use case, then say using custom enums with `ToString` implementation. As such I think that common cases should be as simple and straightforward as possible.

### Proposal
 - Accept bare `bool`, `&str`, `String`, `&String`, `AsRef<str>` and maybe `Cow<str>`
 - One can implement `From<CustomType> for AtValue` for their own custom types.
 - Improve `attrs!` macro so it could accept key without a value:
```rust
attrs!{At::Class => "foo", At::Hidden}
```
- `AtValue::None` use case should be handled by value-less attr key
- User should use `AtValue::*` in his code only as a last resort

### DIFF
--- a/examples/todomvc/src/lib.rs
+++ b/examples/todomvc/src/lib.rs
@@ -226,7 +226,7 @@ fn todo_item(item: &Todo, posit: usize, edit_text: &str) -> Node<Msg> {
                 attrs! {
                    At::Class => "toggle",
                    At::Type => "checkbox",
-                   At::Checked => item.completed.as_at_value()
+                   At::Checked => item.completed,
                 },
                 simple_ev(Ev::Click, Msg::Toggle(posit))
             ],
@@ -317,7 +317,7 @@ fn view(model: &Model) -> impl View<Msg> {
             input![
                 attrs! {
                     At::Id => "toggle-all"; At::Class => "toggle-all"; At::Type => "checkbox",
-                    At::Checked => (model.active_count() == 0).as_at_value(),
+                    At::Checked => model.active_count() == 0,
                 },
                 simple_ev(Ev::Click, Msg::ToggleAll)
             ],
--- a/src/dom_types.rs
+++ b/src/dom_types.rs
@@ -358,6 +358,10 @@ make_attrs! {
 }
 
 /// A thinly-wrapped `HashMap` holding DOM attributes
+///
+/// Attributes can be either String or Boolean. If attribute is
+/// boolean and is true, it should have value of empty string. If it's
+/// false, it should be omitted completely.
 #[derive(Clone, Debug, PartialEq)]
 pub struct Attrs {
     // We use an IndexMap instead of HashMap here, and in Style, to preserve order.
@@ -371,9 +375,9 @@ impl fmt::Display for Attrs {
             .vals
             .iter()
             .filter_map(|(k, v)| match v {
-                AtValue::Ignored => None,
-                AtValue::None => Some(k.as_str().to_string()),
-                AtValue::Some(value) => Some(format!("{}=\"{}\"", k.as_str(), value)),
+                AtValue::Bool(false) => None,
+                AtValue::Bool(true) | AtValue::Empty => Some(k.as_str().to_string()),
+                AtValue::String(value) => Some(format!("{}=\"{}\"", k.as_str(), value)),
             })
             .collect::<Vec<_>>()
             .join(" ");
@@ -409,7 +413,7 @@ impl Attrs {
     pub fn add_multiple(&mut self, key: At, items: &[&str]) {
         self.add(
             key,
-            &items
+            items
                 .iter()
                 .filter_map(|item| {
                     if item.is_empty() {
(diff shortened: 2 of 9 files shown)


David-OConnor:
The behavior used to be similar to this, where bools, numbers etc would auto-coerce. This implementation is better than how it was, but we need to look at the reasons for that change, and if this gets around them.

TatriX:
Updated PR comment with a proposal and motivation. Please share your thoughts :)

TatriX:
Another approach would be
```rust
attrs!{ At::Class("foo"), At::Hidden}
attrs!{ At::Id(String::from("bar"), if state.hidden { At::Hidden } else { () } }
```

David-OConnor:
Original PR: https://github.com/seed-rs/seed/pull/186

TatriX:
> Another approach would be
> 
> ```rust
> attrs!{ At::Class("foo"), At::Hidden, At::Checked(true) }
> attrs!{ At::Id(String::from("bar")), if state.hidden { At::Hidden } else { () } }
> ```

I think this can be a better alternative as we can make API more type safe.
For example we can verify values for attributes, thus enforcing `bool` for certain attributes etc.

MartinKavik:
See issue https://github.com/seed-rs/seed/issues/261 + especially https://github.com/seed-rs/seed/issues/261#issuecomment-546629351 & https://github.com/seed-rs/seed/issues/261#issuecomment-555138892.

I'll dive into the propsal deeply maybe today or tomorrow.

MartinKavik:
Summary of all options.
Correct me if something is wrong (not possible in Rust, typos, I don't understand options from proposal correctly, etc.). Or add your ideas.

---

**Expected output**

```html
<div class="foo" disabled autocomplete="off" autofocus></div>
```

**Use & Variables**

```rust
use AtValue::{Ignored, None};  // conflict with `Option::None`?

let disabled: bool = true;
let autocomplete: bool = false; 
```

**Options:**

1. Current API
```rust
div![
    attrs!{
        At::Class => "foo",
        At::Disabled => disabled.as_at_value(),
        At::Autocomplete => if autocomplete { Ignored } else { "off".into() },
        At::Autofocus => None,
    }
]
```

1.2 Current API - with fixed implementation
```rust
div![
    attrs!{
        At::Class => "foo",
        At::Disabled => disabled,
        At::Autocomplete => if autocomplete { Ignored } else { "off".into() },
        At::Autofocus => None,
    }
]
```

---

2. `()` + `At::WithoutValue` + `Attr => value`
```rust
div![
    attrs!{
        At::Class => "foo",
        if disabled { At::Disabled } else { () },   // is it possible?
        if autocomplete { () } else { At::Autocomplete => "off"  },
        At::Autofocus,
    }
]
```
---

3. `()` + `At::WithoutValue` + `At::WithValue(value)`
```rust
div![
    attrs!{
        At::Class("foo"),
        if disabled { At::Disabled } else { () },
        if autocomplete { () } else { At::Autocomplete(String::from("off"))  },
        At::Autofocus,
    }
]
```

---

4. `()` + `At::function(value)`
```rust
div![
    attrs!{
        At::class("foo"),
        if disabled { At::disabled } else { () },
        if autocomplete { () } else { At::autocomplete("off")  },
        At::autofocus,
    }
]
```

TatriX:
I also started thinking about getting rid of  `attrs!` macro:
```rust
div![
    Attrs::from(&[
        Attr::class("foo"),
        Attr::disabled(disabled),
        Attr::autocomplete(autocomplete),
        Attr::autofocus,
    ]),
];

div![
    &[foo_attr(), bar_attr()].iter().collect::<Attrs>()
}

```

Instead of having `At` and `AtValue` we can have `enum Attr` and `Attrs` can be a simple `Vec` newtype.
```rust
enum attr {
  AutoComplete(bool),
  Disabled(bool),
  FormMethod(FormMethod),
  ...
}
impl Attr {
   fn render(self) -> Option<String> {
      match self {
         Attr::AutoComplete(value) => if value { Some("on") } else { Some("off") },
         Attr::Disabled(value) => if value { Some("") } else { None },
         Attr::FormMethod(method) => method.render(),
        ....
      }
   }
}
```
We can take [data](https://github.com/tawesoft/html5spec/blob/master/spec-json/attributes.json) from html spec and generate `Attr` enum. Spec has information about possible keywords for attrs so we can determine how to render them.
I haven't found yet a way to easily represent attrs with more then 2 possible values, but something like this could work:
```rust
Attr::FormMethod(FormMethod::POST)
```

MartinKavik:
It looks like we're heading to typed attributes and typed values. It was [planned](https://github.com/seed-rs/seed/issues/211#issuecomment-526911813) to do it for CSS and then for other entities.

I like the non-`attrs!` approach because it wouldn't be a big breaking change and it would be more compatible with IDE autocompletes.

So I suggest to create a new issue like _"Typed attributes - API design"_ and close this PR because it looks like the implementation and API will be different. What do you think?

TatriX:
Or we can rename this one. It's up to you.

MartinKavik:
New issue: https://github.com/seed-rs/seed/issues/312.

PR #314, opened 2019-12-15
`Vec<Attr>` and `Vec<Style>` in view macros

David-OConnor:
Could probably expand this to support events, and maps (or filters? Don't remember why we don't have that for child elements) of them.

Example:
```rust
let mut styles = vec![style!{St::Cursor => "pointer"}];
if a {
    styles.push(style!{St::Color => "black"});
}

div![ styles, "text"]
```

Reasoning: Adds more flexibility to the view syntax.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Changelog
 
+[unreleased]
+
 ## v0.5.0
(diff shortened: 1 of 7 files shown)


David-OConnor:
Closing in favor of new one post latest refactor

Issue #128, opened 2019-06-06
Implement UpdateEl for iterators


theduke:
I noticed that `UpdateEl` is not implemented on iterators.

Would be great to add an implementation.
That saves me a annoying collect.

```rust
div![
  // Want: 
  items.iter().map(render_item)
  // Required now:
  items.iter().map(render_item).collect::<Vec<_>>()
]
```

David-OConnor:
Agree.

David-OConnor:
I'm not sure if it's possible to differentiate the type of an `Iterator`. Getting conflicting requirements issues, where the compiler can't tell if we should add `El`s, `Listener`s, or text, due to all of them being iterators.

MartinKavik:
The only way how I can compile:
```rust
fn render_item(number: &i32) -> El<Msg> {
    div![ format!("{}", number) ]
}
fn view(_: &Model) -> El<Msg> {
    let items = vec![1,2,3];
    div![
      items.iter().map(render_item)
    ]
}
```
Is to implement `UpdateEl` for `Map`:
```rust
impl<Ms, T, U> UpdateEl<El<Ms>> for std::iter::Map<T, U>
    where
        T: Iterator,
        U: FnMut(T::Item)->El<Ms>
{
    fn update(self, el: &mut El<Ms>) {
        el.children.append(&mut self.collect());
    }
}
```
But I don't know how to make it more generic, so it's not very useful in the current form.

MartinKavik:
I've exhausted all my ideas.
@theduke Do you have an advice how to implement it? Otherwise I recommend to close it.

David-OConnor:
Let's leave open for now. I don't have a fix now, and suspect their may not be one... But this would be a syntax improvent. Maybe some macro magic we haven't thought of?

MartinKavik:
You motivated me to do some more investigation.

The most generic implementation for iterators I can come up with:
```rust
impl<Ms, I, U, F> UpdateEl<El<Ms>> for std::iter::Map<I, F>
    where
        I: Iterator,
        U: UpdateEl<El<Ms>>,
        F: FnMut(I::Item)->U
{
    fn update(self, el: &mut El<Ms>) {
        self.for_each(|item|item.update(el));
    }
}
```
It let us to use `iter().map(f)` when `f` maps to something that implements `UpdateEl<El<Ms>>`.
So it should cover the most used case when you want to iterate over a list of something from model and render its items.
(I tested it only with some examples so there can be some problems.)

---

[...]

PR #315, opened 2019-12-15
Added support for Vec<Attr> and Vec<Style> in view macros

David-OConnor:
Could probably expand this to support events, and maps (or filters? Don't remember why we don't have that for child elements) of them.

Example:
```rust
let mut styles = vec![style!{St::Cursor => "pointer"}];
if a {
    styles.push(style!{St::Color => "black"});
}

div![ styles, "text"]
```

Reasoning: Adds more flexibility to the view syntax.


MartinKavik:
When do you need it? (i.e. when the current API isn't enough?).

David-OConnor:
If you want to conditionally build the style. There are other ways to currently do this (`.merge()`, `add` and adding multiple `Style`s to the macro), but this may be more natural in some cases, and is consistent with the composable view syntax.

MartinKavik:
I'd rather have one way to build styles (and everything else) and don't extend public API, but if you think some people want it, I agree.
Please add / modify an example to demonstrate it in action - so users know how to use it and we know that it works. 
Add add `.` at the end of the new line in changelog, please 😃

David-OConnor:
Having only one way to do things, in regards to passing collections vs individual items in the view macros, is inconsistent with the way the view syntax is set up For example, there are already 3 ways to combine styles, and you can add children via `.add_child()`, passing a child, passing a vec of children, or passing a map of children. This is deliberate, and one of the things I think is nicer in Seed than other (eg html-like, or pure-fn) view syntaxes.

MartinKavik:
Ok, add those examples pls and let's merge it :)

David-OConnor:
https://github.com/seed-rs/seed-rs.org/pull/13/files

David-OConnor:
Added a re-export of `Style`, `Attrs`, and `Listener`; after the recent refactor, they were buried pretty deep. Can now be imported as `seed::Attrs` etc, for example, in type signatures.

src/lib.rs
```diff
@@ -14,13 +14,14 @@ pub use crate::{
...
-    browser::service::fetch::{Method, Request},
+    browser::service::fetch::{Method, Request, ResponseDataResult, ResponseResult},
     browser::service::routing::push_route,
     browser::url::Url,
     browser::util::{
         self, body, canvas, canvas_context_2d, cookies, document, error, history, html_document,
         log, update, window,
     },
+    virtual_dom::{attrs::Attrs, listener::Listener, style::Style},
 };
```
MartinKavik: 1. `virtual_dom::{Attrs, Listener, Style},` should be enough thanks to reimports.
2. If we need `Attrs` and `Style` I think it should be also added in the `prelude`.
3. How we'll decide what should be in `pub use crate` and what in `prelude`?

src/virtual_dom/update_el.rs
```diff
@@ -111,3 +143,58 @@ where
...
+//impl<Ms, I, U, F> UpdateEl<El<Ms>> for std::iter::Map<I, F>
+//where
+//    I: Iterator,
+//    U: UpdateEl<&Style>,
+//    F: FnMut(I::Item) -> U,
+//{
+//    fn update(self, el: &mut El<Ms>) {
+//        self.for_each(|item| item.update(el));
+//    }
+//}
```
MartinKavik: ?

src/lib.rs
David-OConnor: What do you think about these specific things, on prelude vice crate? It's a balance between making things easy to get to, and not adding too many things to the global namespace. I think these are probably not common enough to be in the prelude; ie the main use I've found for them is in fn signatures.

src/virtual_dom/update_el.rs
David-OConnor: A failed attempt to make `update` work on `Filter`, like we have with `Map`, ie so you don't have to add `collect()`. I think it'd be nice to have, but can't get it working.

src/lib.rs
MartinKavik: Well, with the common import `use seed::{prelude::*, *}` we can maybe remove `pub use crate::{`.
And I haven't used `Attrs` and `Style` at all in `kavik.cz` or `RealWorld ex`, only `Listener` in `window_events`.
I think we can wait once there are more Seed apps, then search through their code and change `prelude` / `pub crate` accordingly as the part of a bigger `lib.rs` refactor. It will be a breaking change but easily fixable one.

src/virtual_dom/update_el.rs
MartinKavik: https://github.com/seed-rs/seed/issues/128#issuecomment-570337605

PR #321, opened 2019-12-20
Cleaned up Clippy checks

David-OConnor:
The `must_use` lint seems a little... aggressive. Disabled it for the main repo. CI is still failing for examples. I don't want to add either this tag, or the allow to examples; maybe we can use normal `clippy` instead of `clippy::pedantic` in the examples.


David-OConnor:
CI now passes. I think the blanket allow may be the best approach here; `pedantic` lints are designed to be aggressive, and may not always be appropriate.

Issue #316, opened 2019-12-15
listeners disapper when cloneing Node or El


MuhannadAlrusayni:
I'm trying to store nodes in structs and use them later in view.

use case:
``` rust
    #[derive(Clone, Debug)]
    pub enum Msg {
        Home,
        AboutUs
    }

    #[derive(Debug)]
    pub struct Flexbox {
        items: Vec<Node<Msg>>,
    }

    fn view_flexbox(flexbox: &Flexbox) -> Node<Msg> {
        div![
            flexbox.items.clone()
        ]
    }

    #[test]
    fn test_listeners() {
        let items = vec![
            button![simple_ev(Ev::Click, Msg::Home), "Home"],
            button![simple_ev(Ev::Click, Msg::AboutUs), "About us"]
        ];

        let flexbox = Flexbox {items};
        // this panic show that flexbox items have listeners
        panic!("{:#?}", flexbox);
        let node = view_flexbox(&flexbox);
        // this panic show that node doesn't children doesn't have listeners
        panic!("{:#?}", node);
    }

```

It seems listeners are not cloned with the node/el:

https://github.com/seed-rs/seed/blob/60d7de4386b18e95091d5f16716664f77cfed003/src/virtual_dom/node/el.rs#L206-L221

Is there way to fix this ?

MartinKavik:
I don't know, @David-OConnor ?
However I've added it to my todo-list - https://github.com/seed-rs/seed/issues/293. I plan to do a bigger update for virtual DOM in January.
Or you can try to fix it.

MuhannadAlrusayni:
I don't think I can do it, and if I try I need to understand the internal code for seed first and this would need non trivial time which I don't have right now.

MuhannadAlrusayni:
Yew seems to have similar issue, and they already have draft PR https://github.com/yewstack/yew/pull/786 , it might be helpful to get some inspiration from them.

MartinKavik:
@MuhannadAlrusayni thanks. I'll be working on it in January.
Closing in favor of #293

MuhannadAlrusayni:
I got time to solve this issue, but I need some assistance :D.

So the problem is that `Listener<Ms>` doesn't implement `Clone`, it also contains `EventHandler<Ms>` and `Closure<_>` both don't implement `Clone` too.

[...]

Issue #293, opened 2019-11-21
VirtualDOM related issues / todos


MartinKavik:
- Event listener patching - https://github.com/seed-rs/seed/issues/138#issuecomment-554961733
- Refs - https://github.com/seed-rs/seed/issues/115#issuecomment-557036133
- VirtualNode - https://github.com/seed-rs/seed/issues/180#issue-465733737

PR #323, opened 2019-12-26
impl `Clone` for `Listener` and `EventHandler`

MuhannadAlrusayni:
Solve https://github.com/seed-rs/seed/issues/316

### DIFF
--- a/src/virtual_dom/listener.rs
+++ b/src/virtual_dom/listener.rs
@@ -24,13 +24,6 @@ impl<Ms> PartialEq for EventHandler<Ms> {
     }
 }
 
-// TODO: try to print Ms too, maybe by restricting Ms: Debug ?
-impl<Ms> fmt::Debug for EventHandler<Ms> {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.write_str("EventHandler<_>")
-    }
-}
-
 impl<Ms> EventHandler<Ms> {
     pub fn call(&self, event: web_sys::Event) -> Ms {
         (self.0)(event)
--- a/src/virtual_dom/listener.rs
+++ b/src/virtual_dom/listener.rs
@@ -1,11 +1,56 @@
 use super::Ev;
 use crate::app::MessageMapper;
 use crate::browser::{dom::lifecycle_hooks::fmt_hook_fn, util::ClosureNew};
-use enclose::enclose;
-use std::{fmt, mem};
+use std::{fmt, mem, rc::Rc};
 use wasm_bindgen::{closure::Closure, JsCast};
 
-type EventHandler<Ms> = Box<dyn FnMut(web_sys::Event) -> Ms>;
+pub struct EventHandler<Ms>(Rc<dyn Fn(web_sys::Event) -> Ms>);
+
+impl<Ms, F: Fn(web_sys::Event) -> Ms + 'static> From<F> for EventHandler<Ms> {
+    fn from(func: F) -> Self {
+        EventHandler(Rc::new(func))
+    }
+}
+
+impl<Ms> Clone for EventHandler<Ms> {
+    fn clone(&self) -> Self {
+        EventHandler(self.0.clone())
+    }
+}
+
+impl<Ms> PartialEq for EventHandler<Ms> {
+    fn eq(&self, other: &EventHandler<Ms>) -> bool {
+        Rc::ptr_eq(&self.0, &other.0)
+    }
+}
+
+// TODO: try to print Ms too, maybe by restricting Ms: Debug ?
+impl<Ms> fmt::Debug for EventHandler<Ms> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str("EventHandler<_>")
+    }
+}
+
+impl<Ms> EventHandler<Ms> {
+    pub fn call(&self, event: web_sys::Event) -> Ms {
+        (self.0)(event)
+    }
+}
+
+impl<Ms: 'static, OtherMs: 'static> MessageMapper<Ms, OtherMs> for EventHandler<Ms> {
+    type SelfWithOtherMs = EventHandler<OtherMs>;
+    fn map_msg(
+        self,
+        msg_mapper: impl FnOnce(Ms) -> OtherMs + 'static + Clone,
+    ) -> EventHandler<OtherMs> {
+        let orignal_handler = self;
+        let new_handler = move |event| {
+            let msg = orignal_handler.call(event);
+            (msg_mapper.clone())(msg)
+        };
+        EventHandler(Rc::new(new_handler))
+    }
+}
 
 #[derive(Clone, Copy, Debug, PartialEq)]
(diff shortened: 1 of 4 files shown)


MuhannadAlrusayni:
Note that `cargo make verify` doesn't compile successfully because of some clippy errors that didn't occur from this PR, `cargo build` compile without errors.

I have tried this PR from old seed project I have and it works fine. 

I didn't try to compile examples, will do if this PR will be accepted.

David-OConnor:
Looks good. This functionality was missing from the beginning. I need to do some testing first to make sure removing the `Box`s didn't break something subtle. Listeners / event-handling's always been the trickiest part of the code.

src/virtual_dom/listener.rs
```diff
@@ -1,11 +1,56 @@
...
+        Rc::ptr_eq(&self.0, &other.0)
+    }
+}
+
+// TODO: try to print Ms too, maybe by restricting Ms: Debug ?
+impl<Ms> fmt::Debug for EventHandler<Ms> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str("EventHandler<_>")
+    }
+}
```
MartinKavik: Is `fmt::Debug` impl for `EventHandler` necessary? And we shouldn't restrict `Ms` to `Debug`. It should be possible to pass `EventHandler` into `log!` if you explicitly restrict `Ms: Debug` in your application (i.e. in the root) if you are on `nightly` Rust - see https://github.com/seed-rs/seed/pull/234/files

src/virtual_dom/node.rs
```diff
@@ -19,7 +19,7 @@ pub enum Node<Ms: 'static> {
     Empty,
 }
 
-impl<Ms: 'static> Clone for Node<Ms> {
+impl<Ms: Clone + 'static> Clone for Node<Ms> {
```
MartinKavik: I hoped we can remove all `Clone` restrictions for `Ms` because it force us to do some ugly things like wrapping `Msg` variants to `Rc` in some applications.

src/virtual_dom/node/el.rs
```diff
@@ -205,15 +205,15 @@ impl<Ms> El<Ms> {
 
 /// Allow the user to clone their Els. Note that there's no easy way to clone the
 /// closures within listeners or lifestyle hooks, so we omit them.
-impl<Ms> Clone for El<Ms> {
+impl<Ms: Clone> Clone for El<Ms> {
```
MartinKavik: I hoped we can remove all `Clone` restrictions for `Ms` because it force us to do some ugly things like wrapping `Msg` variants to `Rc` in some applications.

src/virtual_dom/listener.rs
```diff
@@ -37,6 +82,21 @@ pub struct Listener<Ms> {
     message: Option<Ms>,
 }
 
+impl<Ms: Clone> Clone for Listener<Ms> {
+    fn clone(&self) -> Self {
+        Self {
+            trigger: self.trigger,
+            handler: self.handler.clone(),
+            // closure shouldn't be cloned since the new Listener isn't related to this Listener
+            closure: None,
```
MartinKavik: Could you explain it a little bit more, please? I don't fully understand it, thanks.

MartinKavik:
> Looks good. This functionality was missing from the beginning. I need to do testing first to make sure removing the Boxs didn't break something subtle. Listeners / event-handling's always been the most finicky part of the codebase.

Good comment. I want to rewrite event-handling a bit as a part of https://github.com/seed-rs/seed/issues/293, so I think we don't have to spend very much time with testing because it's possible I'll change the code in this PR (some days/weeks after merge to `master`).

David-OConnor:
I'm cool with merging if you are.

MartinKavik:
Let's wait a little bit for @MuhannadAlrusayni comments in the review, but yeah, we can merge it then.

src/virtual_dom/listener.rs
MuhannadAlrusayni: > Is fmt::Debug impl for EventHandler necessary?

Good point, some why I thought this was needed... we can remove this if it's not needed.

src/virtual_dom/node.rs
MuhannadAlrusayni: This is needed to clone `Listener<Ms>` who owns `message: Option<Ms>`.

We can remove `message: Option<Ms>` this way we can clone `Listener<Ms>` thus we can clone `Node<Ms>` without restricting `Ms: Clone.`.

if I understand correctly `message: Option<Ms>` is only used in `impl<Ms> PartialEq for Listener<Ms>` which I think it isn't important and it isn't complete too.

src/virtual_dom/node/el.rs
MuhannadAlrusayni: Same stroy for `Node<Ms>`

src/virtual_dom/listener.rs
MuhannadAlrusayni: > So the problem is that Listener<Ms> doesn't implement Clone, it also contains EventHandler<Ms> and Closure<_> both don't implement Clone too.
> 
> First let's talk about EventHandler<Ms> which contains a closure Box<dyn FnMut(web_sys::Event) -> Ms>, this closure will be used to create Closure<_> in attach function.
> 
> Then we have Closure<_> which get created in attach function, and passed to add_event_listener_with_callback which add event listener in js world, Listener<Ms> stores this closure so it can detach it later with detach function.
> 
> If I understood correctly, EventHandler<Ms> must be cloned when we clone Listener<Ms>, while Closure<_> shouldn't be cloned when we clone Listener<Ms>.
> 
> So we have to implement Clone for EventHandler<Ms>, one way to do that is by wrap type that contain Rc. Then implement Clone for Listener<Ms>.
> 
> Please correct me if I miss something, if I my understood is correct I can open PR for this.

src/virtual_dom/listener.rs
MartinKavik: I think that `log!` support for implicit `Debug` objects on nightly is enough. 
Remove it, if you want or I'll probably remove it during VDOM refactor.

src/virtual_dom/node.rs
MartinKavik: > if I understand correctly message: Option<Ms> is only used in impl<Ms> PartialEq for Listener<Ms> which I think it isn't important and it isn't complete too.

It will be probably a part of my VDOM refactor so it's possible I'll resolve it. So we can leave it as it is for now.

src/virtual_dom/listener.rs
MartinKavik: I think you know about it more than me now and it make sense. And it looks like another VDOM part which I want to update. Let's leave it as is for now.

MartinKavik:
@MuhannadAlrusayni Do you want to change something or can I merge it?

MuhannadAlrusayni:
I have removed `impl<Ms> Debug for EventHandler<Ms>`, we can merge now.

MartinKavik:
It looks like we forgot to update changelog.

PR #326, opened 2019-12-29
0.5.1

David-OConnor:
Includes a few recently changes, but is mainly to rebuild the docs.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,7 +1,11 @@
 //! See readme for details.
 
 //#![deny(missing_docs)]
-#![allow(clippy::use_self, clippy::single_match_else)]
+#![allow(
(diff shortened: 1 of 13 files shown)


src/lib.rs
```diff
@@ -14,8 +14,10 @@ pub use crate::{
     browser::dom::cast::{
         to_html_el, to_input, to_kbevent, to_mouse_event, to_select, to_textarea,
     },
+    browser::service::fetch,
     browser::service::fetch::{Method, Request, ResponseDataResult, ResponseResult},
```
MartinKavik: `browser::service::fetch::{self, Method, Request, ResponseDataResult, ResponseResult}` could be cleaner.

PR #337, opened 2020-01-21
impl Clone manually for EventHandler<Ms>

MuhannadAlrusayni:
the current implementation does require `Ms` to impl `Clone` :D
https://doc.rust-lang.org/stable/std/clone/trait.Clone.html#derivable

### DIFF
--- a/src/virtual_dom/event_handler_manager/event_handler.rs
+++ b/src/virtual_dom/event_handler_manager/event_handler.rs
@@ -2,11 +2,19 @@ use crate::app::MessageMapper;
 use crate::virtual_dom::Ev;
 use std::{fmt, rc::Rc};
 
-#[derive(Clone)]
 /// `EventHandler`s are called by DOM event listeners with the same trigger (an event to listen to).
 pub struct EventHandler<Ms> {
     pub trigger: Ev,
     pub callback: Rc<dyn Fn(web_sys::Event) -> Ms>,
+}
+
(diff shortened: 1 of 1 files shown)


MartinKavik:
Yeah, it's annoying because of the current Rust `#[derive(Clone)]` implementation.
Add line like this
https://github.com/seed-rs/seed/blob/5d94d966fec97d7b4dfa841259f87fa5dbcb831a/src/virtual_dom/node/el.rs#L34
if you want.
Then we can merge it.

MartinKavik:
Yeah, I didn't notice your link :)

src/virtual_dom/event_handler_manager/event_handler.rs
```diff
@@ -2,13 +2,21 @@ use crate::app::MessageMapper;
...
 pub struct EventHandler<Ms> {
     pub trigger: Ev,
     pub callback: Rc<dyn Fn(web_sys::Event) -> Ms>,
 }
 
+impl<Ms> Clone for EventHandler<Ms> {
+    fn clone(&self) -> Self {
+        Self {
+            trigger: self.trigger.clone(),
+            callback: self.callback.clone(),
```
MartinKavik: `callback: Rc::clone(..)` pls

MartinKavik:
yes, I'll merge it as soon as tests pass

PR #352, opened 2020-02-13
Refactoring: replace `Result<T, T>` with just `T`

TatriX:
There is no point in having `Result<T, T>` instead of just `T`.
It seems that this happened due to migration from `futures 0.1` to `std::future`.


MartinKavik:
Note: Just make sure, please, that doc comments with examples are updated.

PR #353, opened 2020-02-15
New Fetch API

TatriX:


src/browser/fetch.rs
```diff
@@ -0,0 +1,128 @@
...
+    fn from(string: String) -> Resource<'static> {
+        Resource::String(Cow::from(string))
+    }
+}
+
+impl From<Url> for Resource<'_> {
+    fn from(url: Url) -> Resource<'static> {
+        Resource::String(Cow::from(url.to_string()))
+    }
+}
```
MartinKavik: I think we can remove this if `Url` automatically implements `Into<Cow>` (with `Display`). (If we used the code from my previous comment (`impl <'a, T> From<T> for Resource<'a> where ...`).

src/browser/fetch.rs
```diff
@@ -0,0 +1,128 @@
...
+
+/// The fetch functions is a main entry point of the Fetch API.
+///
+/// It start the process of fetching a resource from the network,
+/// returning a future which is fulfilled once the response is
+/// available. The future resolves to the Response object representing
+/// the response to your request. The promise does not reject on HTTP
+/// errors — it only rejects on network errors. You must use then
+/// handlers to check for HTTP errors.
+pub async fn fetch<'a>(resourse: impl Into<Resource<'a>>) -> Result<Response> {
```
MartinKavik: Why we need `Resource`? Isn't `impl Into<Request<'a>` enough?
We can implement 
```rust
impl <'a, T> From<T> for Request<'a>
    where
        T: Into<Cow<'a, str>>
```
and it should cover all cases (?)

MartinKavik: resour **c** e

src/browser/fetch/header.rs
```diff
@@ -0,0 +1,54 @@
...
+}
+
+impl<'a> IntoIterator for Headers<'a> {
+    type Item = Header<'a>;
+    type IntoIter = std::vec::IntoIter<Self::Item>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.0.into_iter()
+    }
+}
```
MartinKavik: Why we need `Headers` struct? And why it contains `Vec`?
Can we replace it with `BTreeMap<Header>`?
  - It already implements iterators.
  - We don't need to use weird combination of `self.0.retain` + `push`.
  - It would be ordered in logs.

src/browser/fetch/header.rs
```diff
@@ -0,0 +1,54 @@
...
+    custom("Authorization", format!("Bearer {}", value.into()))
+}
+
+/// Create custom header.
+pub fn custom<'a>(name: impl Into<Cow<'a, str>>, value: impl Into<Cow<'a, str>>) -> Header<'a> {
+    Header {
+        name: name.into(),
+        value: value.into(),
+    }
+}
```
MartinKavik: Why are they module functions and not `Header` methods?

src/browser/fetch/method.rs
```diff
@@ -0,0 +1,37 @@
+/// HTTP Method types.
+///
+/// [https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods)
```
MartinKavik: ```/// [MDN reference](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods)``` should be enough

src/browser/fetch/method.rs
```diff
@@ -0,0 +1,37 @@
...
+        match self {
+            Method::Get => "GET",
+            Method::Head => "HEAD",
+            Method::Post => "POST",
+            Method::Put => "PUT",
+            Method::Delete => "DELETE",
+            Method::Connect => "CONNECT",
+            Method::Options => "OPTIONS",
+            Method::Trace => "TRACE",
+            Method::Patch => "PATCH",
```
MartinKavik: `Method` vs `Self`

src/browser/fetch/request.rs
```diff
@@ -0,0 +1,282 @@
...
+
+    /// Set HTTP method. Default method is `GET`.
+    pub const fn method(mut self, method: Method) -> Self {
+        self.method = method;
+        self
+    }
+
+    /// Set request body to provided `JsValue`. Consider using `json` or `text` methods instead.
+    ///
+    /// Note that a request using the GET or HEAD method cannot have a body.
```
MartinKavik: Add a note or `/// # Panics` block, please, so users know that it panics when you set body and use GET or HEAD.

src/browser/fetch/request.rs
```diff
@@ -0,0 +1,282 @@
...
+    /// Set request body by JSON encoding provided data.
+    /// It will also set `Content-Type` header to `application/json; charset=utf-8`.
+    ///
+    /// # Errors
+    ///
+    /// This method can fail if JSON serialization fail. It will then
+    /// return `FetchError::SerdeError`.
+    pub fn json<T: Serialize + ?Sized>(mut self, data: &T) -> Result<Self> {
+        self.headers
+            .set(header::content_type("application/json; charset=utf-8"));
```
MartinKavik: Or I think we can use `self.header(..`.
(The same for other methods.)

src/browser/fetch/request.rs
```diff
@@ -0,0 +1,282 @@
...
+    /// Set request mode.
+    /// It can either be `cors`, `no-cors`, `same-origin`, or `navigate`.
+    /// The default is `cors`.
+    pub const fn mode(mut self, mode: web_sys::RequestMode) -> Self {
+        self.mode = Some(mode);
+        self
+    }
+
+    /// Set request credentials.
+    /// It can either be `omit`, `same-origin`, or `include`.
```
MartinKavik: Could you add `# Example` block?
(The same for other methods, where it's not clear how to use enum of values from the description)

src/browser/fetch/request.rs
```diff
@@ -0,0 +1,282 @@
...
+    /// The default is `follow`.
+    pub const fn redirect(mut self, redirect: web_sys::RequestRedirect) -> Self {
+        self.redirect = Some(redirect);
+        self
+    }
+
+    /// Set request referrer.
+    /// It can be either `referrer`, `client`, or a `URL`.
+    /// The default is `about:client`.
+    pub fn referrer(mut self, referrer: String) -> Self {
```
MartinKavik: Why `String`? I assume that user would use probably `&'static str`.
(The same for other methods which have `String` as input parameter type.)

src/browser/fetch/request.rs
```diff
@@ -0,0 +1,282 @@
...
+                Timeout::new(*timeout, move || abort_controller.abort()),
+            );
+        }
+
+        // controller
+        // https://developer.mozilla.org/en-US/docs/Web/API/AbortController/signal
+        init.signal(Some(&request.controller.abort_controller.signal()));
+
+        // It seems that the only reason why Request constructor can
+        // fail is when Url contains credentials.  I assume that this
```
MartinKavik: Could you provide more details? And maybe add a short note to `expect`'s text?

src/browser/fetch/response.rs
```diff
@@ -0,0 +1,49 @@
...
+/// To get one you need to use [`fetch`](./fn.fetch.html) function.
+///
+/// [MDN reference](https://developer.mozilla.org/en-US/docs/Web/API/Response)
+pub struct Response {
+    pub(crate) raw_response: web_sys::Response,
+}
+
+impl Response {
+    /// Get a `String` from response body.
+    pub async fn text(self) -> Result<String> {
```
MartinKavik: Please add `# Errors` block.
(The same for other methods)

examples/fetch/src/post.rs
```diff
@@ -0,0 +1,66 @@
...
+            let request = Request::new("/")
+                .method(Method::Post)
+                .header(header::custom("Accept-Language", "Rust"))
+                .header(header::custom("Authorization", format!("Basic {}", token)))
+                .json(&model.form)
+                .expect("Serialization failed");
+
+            orders.perform_cmd(async {
+                let response = fetch(request).await.expect("HTTP request failed");
+                assert!(response.status().is_ok());
```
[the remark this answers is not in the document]
TatriX: Well assertions is not about tests, they are about invariants. In this case this means "I expect that status is ok, otherwise panic", i.e. [fail fast](https://en.wikipedia.org/wiki/Fail-fast)
As a side note, `assert` will still be present in the release build, compared to `debug_assert`.

examples/fetch/src/simple.rs
```diff
@@ -0,0 +1,48 @@
...
+    Received(User),
+}
+
+pub fn update(msg: Msg, model: &mut Model, orders: &mut impl Orders<Msg>) {
+    match msg {
+        Msg::Fetch => {
+            orders.skip(); // No need to rerender
+            orders.perform_cmd(async {
+                let response = fetch("user.json").await.expect("HTTP request failed");
+                let user = response
```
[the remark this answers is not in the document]
TatriX: I made this example as fast and simple and short as possible. Amusing this will be used as a first experimental iteration in the code evolution.

examples/server_interaction/src/lib.rs
```diff
@@ -58,7 +58,11 @@ impl Default for Model {
 // ------ ------
 
 fn after_mount(_: Url, orders: &mut impl Orders<Msg>) -> AfterMount<Model> {
-    orders.perform_cmd(fetch_repository_info());
+    orders.perform_cmd(
+        fetch(REPOSITORY_URL)
+            .and_then(Response::json)
```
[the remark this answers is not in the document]
TatriX: Nope, it doesn't. Should we maybe add some helper method/function for this?

src/browser/url.rs
```diff
@@ -60,6 +60,18 @@ impl Url {
...
 
+impl std::fmt::Display for Url {
+    fn fmt(&self, fmt: &mut std::fmt::Formatter) -> std::fmt::Result {
+        let path = self.path.join("/");
+        match (&self.search, &self.hash) {
+            (Some(search), Some(hash)) => write!(fmt, "{}?{}#{}", path, search, hash),
+            (Some(search), None) => write!(fmt, "{}?{}", path, search),
+            (None, Some(hash)) => write!(fmt, "{}#{}", path, hash),
+            _ => write!(fmt, "{}", path),
+        }
```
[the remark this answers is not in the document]
TatriX: Should we then instead of custom made Url object use something like https://github.com/servo/rust-url?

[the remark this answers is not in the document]
TatriX: Ah, right. Found your issue there about wasm size. I think that if we can omit websys call then we should. Why do you think it's error prone? Not that I'm not having the same gut feeling, but still what could go wrong? ;)

src/browser/fetch.rs
```diff
@@ -0,0 +1,128 @@
+//! Fetch API.
+//!
+//! Our version of the Fetch API is based mostly on regular web one.
+//!
+//! There is one entry point: [`fetch`][fetch] function.
+//! It can accept both, String urls as well as [`Request`][request].
```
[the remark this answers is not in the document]
TatriX: I'll use lowercase then, cause it also works with `&str`.

src/browser/fetch.rs
```diff
@@ -0,0 +1,128 @@
...
+pub type Result<T> = std::result::Result<T, FetchError>;
+
+/// The fetch functions is a main entry point of the Fetch API.
+///
+/// It start the process of fetching a resource from the network,
+/// returning a future which is fulfilled once the response is
+/// available. The future resolves to the Response object representing
+/// the response to your request. The promise does not reject on HTTP
+/// errors — it only rejects on network errors. You must use then
+/// handlers to check for HTTP errors.
```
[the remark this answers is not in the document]
TatriX: Heh, I copied that text from mdn I think.

src/browser/fetch.rs
```diff
@@ -0,0 +1,128 @@
...
+pub async fn fetch<'a>(resourse: impl Into<Resource<'a>>) -> Result<Response> {
+    let promise = match resourse.into() {
+        Resource::String(string) => window().fetch_with_str(&string),
+        Resource::Request(request) => window().fetch_with_request(&request.into()),
+    };
+
+    let raw_response = JsFuture::from(promise)
+        .await
+        .map(Into::into)
+        .map_err(|js_value_error| FetchError::DomException(js_value_error.into()))?;
```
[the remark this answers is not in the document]
TatriX: As far as I understand it fails when promise rejects, so on network error. But I'm not sure.

src/browser/fetch.rs
```diff
@@ -0,0 +1,128 @@
...
+    DomException(web_sys::DomException),
+    PromiseError(wasm_bindgen::JsValue),
+}
+
+/// Wrapper for `fetch` function single argument.
+///
+/// Consider using `String` or `Request` instead, because there are
+/// `From` implementations for those types.
+pub enum Resource<'a> {
+    String(Cow<'a, str>),
```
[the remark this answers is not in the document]
TatriX: I based it on terminology from [fetch docs](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch). It's either `USVString` or `Request`.

src/browser/fetch.rs
TatriX: To utilize `window().fetch_with_str(&string)` and `fetch_with_request(&request.into())` so we don't have to construct request.

src/browser/fetch/header.rs
TatriX: Mostly for performance reasons. Headers most of the time are very small, thus using hashmaps/trees for tiny number of entries has a worse performance, despite o(1) vs o(n).

Eventually we will also probably add [append method](https://developer.mozilla.org/en-US/docs/Web/API/Headers/append), so separate struct is desirable.

src/browser/fetch/header.rs
TatriX: No idea tbh :crazy:

src/browser/fetch/method.rs
TatriX: Hm?

examples/fetch/src/simple.rs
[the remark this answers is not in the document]
MartinKavik: Ok

src/browser/url.rs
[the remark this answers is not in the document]
MartinKavik: > Why do you think it's error prone?

Once we add another URL part I can guarantee that we will forget to add it into some `write!` function. And once we want to refactor it - e.g. encode hash - again the same problem. And once we add some order-sensitive part (if it's possible) => `match (&self.search, &self.hash, &self.foo)`.. it will be pain.

> I think that if we can omit websys call then we should.

It sounds like premature optimizations. It does make some sense in patching algorithm; it isn't very important here.

src/browser/fetch/request.rs
TatriX: These methods expose websys, thus they are documented on MDN/websys. Not sure what is the best way to solve this.

src/browser/fetch/request.rs
TatriX: Panic looks like this:
```
package.js:280 panicked at 'fetch: Cannot create request: JsValue(TypeError: Failed to construct 'Request': Request cannot be constructed from a URL that includes credentials
```
so it seems that all relevant info is already there.

src/browser/fetch.rs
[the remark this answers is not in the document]
MartinKavik: Read paragraphs under the title `Terminology` - https://package.elm-lang.org/packages/elm/browser/latest/Browser-Dom. So pls don't use JS docs and terminology directly in Seed..

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 [unreleased]
 - [BREAKING] Changed `perform_cmd` and `fetch` return type to `T` instead of `Result<T, T>`.
```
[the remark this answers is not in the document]
TatriX: Remove.

src/browser/fetch.rs
MartinKavik: I don't know if more complex public API and more Rust code (~ more potential bugs and a little bit bigger file size) is a good trade-off for this performance (?) optimization.

src/browser/fetch/header.rs
MartinKavik: > Eventually we will also probably add append method, so separate struct is desirable.

Then. I would rather modify `Header` - change its field `value` to `values: Vec<..>`.

> Mostly for performance reasons.

We lose some performance in `Header::set` with `Vec` so I don't think there will be a difference. And I think it's again premature optimization.

src/browser/fetch/method.rs
MartinKavik: e.g. `Self::Head => "HEAD"` instead of `Method::Head => "HEAD"`. I'm surprised that Clippy is ok wih it.

CHANGELOG.md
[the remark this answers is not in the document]
TatriX: Changed the text. That was already merged to master as far as I remember.

examples/fetch/src/post.rs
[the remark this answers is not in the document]
TatriX: Essentially it's a placeholder akin to `.unwrap()` in example code.

src/browser/url.rs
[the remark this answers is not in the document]
TatriX: I remade it using `websys::Url` and it still has all issues you mentioned, with added `expect` call.
If you have better ideas on how it should be done, please share them.

src/browser/fetch.rs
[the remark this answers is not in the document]
TatriX: I think number of people coming from Elm should be negligible compared to people from js world. As for terminology text, I have never thought about "real" scrolls, because I connect scrollbars to mouse scroll. Anyways, should I rename it to  `Url` or leave it as is?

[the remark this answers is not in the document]
MartinKavik: I don't want to bring Elm terminology to Seed - the idea was that we should create our own and simple Seed terminology and take inspiration from other places only if it makes sense.

---

I think `String` is too general, I suggest to rename it.

src/browser/fetch/request.rs
MartinKavik: What about:
```rust
/// Set request credentials.
///
/// Variants: `Omit`, `SameOrigin`, `Include`
/// Default: `SameOrigin`
///
/// [MDN reference](https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials)
/// [`web_sys` reference](https://rustwasm.github.io/wasm-bindgen/api/web_sys/enum.RequestCredentials.html)
```

src/browser/fetch/request.rs
MartinKavik: I'm not very happy with it. We are hiding potential error with panic and also breaking `From` trait basic rule. And I can imagine it can fail due to a dozen undocumented reasons in other browsers...

src/browser/fetch/method.rs
MartinKavik: The same for `impl Default`.

src/browser/url.rs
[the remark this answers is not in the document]
rebo: I've had a look at the latest commit code which uses web_sys::Url.

TatriX, when you say it has the same issues do you mean if another URL part is added? 

I think that would affect any solution however I think it is more maintainable and readable with the current commit code below, so i'm happy with this `web_sys::Url` approach:

```rust
        let url = web_sys::Url::new_with_base(&self.path.join("/"), dummy_base_url)
             .expect("cannot create url");
         if let Some(search) = &self.search {
             url.set_search(search);
         }
         if let Some(hash) = &self.hash {
             url.set_hash(hash);
         }
```

src/browser/fetch.rs
```diff
@@ -1,9 +1,9 @@
 //! Fetch API.
 //!
-//! Our version of the Fetch API is based mostly on regular web one.
+//! Seed Fetch API is very similar to the browser [native one][fetch-mdn].
 //!
 //! There is one entry point: [`fetch`][fetch] function.
-//! It can accept both, String urls as well as [`Request`][request].
+//! It can accept both, string urls as well as [`Request`][request].
```
rebo: Minor grammar quibble, no comma needed after "... accept both, "

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
...
+- Added new Fetch API module. See [`seed::browser::fetch`](https://docs.rs/seed/0.6.0/seed/browser/fetch/index.html) (#353)
```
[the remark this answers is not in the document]
MartinKavik: Why is it resolved? Am I overlooking something?

examples/fetch/src/post.rs
```diff
@@ -0,0 +1,83 @@
...
+
+pub fn update(msg: Msg, model: &mut Model, orders: &mut impl Orders<Msg>) {
+    match msg {
+        Msg::NameChanged(name) => model.form.name = name,
+        Msg::Submit => {
+            orders.skip(); // No need to rerender
+
+            let token = "YWxhZGRpbjpvcGVuc2VzYW1l";
+            // Created outside async block because of lifetime reasons
+            // (we can't use reference to `model.from` in async
```
MartinKavik:
```suggestion
            // (we can't use reference to `model.form` in async
```

examples/fetch/src/post.rs
```diff
@@ -0,0 +1,83 @@
...
+            // function).
```
MartinKavik: Is it on the standalone line because of GitHub weird line breaks or Rust formatter?

MartinKavik:
Closing in favor of PR https://github.com/seed-rs/seed/pull/402

Issue #358, opened 2020-02-18
Add `visibilitychange` event


akhilman:
Add `visibilitychange` to window events.
> The visibilitychange event is fired at the document when the content of its tab have become visible or have been hidden.

https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event

MartinKavik:
Event API will be redesigned - https://github.com/seed-rs/seed/issues/331. And variants will be auto-generated from the online database (see https://github.com/seed-rs/seed/issues/312#issuecomment-565962781 if you want to know how it works).
Until the new API is designed & implemented feel free to create a PR with `visibilitychange` or create event from string (aka custom event) in your project.

MartinKavik:
It's in `master`, closing.

PR #360, opened 2020-02-21
Added visibility change event

David-OConnor:
https://github.com/seed-rs/seed/issues/358


MartinKavik:
Update changelog pls

David-OConnor:
I don't think that's required for adding an event. Ie we wouldn't updated it for an auto-populated one.

MartinKavik:
> we wouldn't update it for an auto-populated one

I disagree - there can be breaking changes.
But I agree for this case - it's not really important and it's an addition.

PR #367, opened 2020-02-23
Impl `UpdateEl` for `Option<T>`, `Result<T, E>`, `Vec<T>`

MuhannadAlrusayni:


src/virtual_dom/update_el.rs
```diff
@@ -179,3 +128,82 @@ where
...
+impl<Ms, T, E> UpdateEl<El<Ms>> for Result<T, E>
+where
+    T: UpdateEl<El<Ms>>,
+    E: UpdateEl<El<Ms>>,
+{
+    fn update(self, el: &mut El<Ms>) {
+        match self {
+            Ok(val) => val.update(el),
+            Err(err) => err.update(el),
+        }
```
MartinKavik: I suggest to change it to:
```rust
if let Ok(val) = self {
    val.update(el);
}
```
- So it's prepared for future `UpdateEl` implementation for `IntoInterator`.
- Users are not limited by `UpdateEl` restriction for `err`.
- My intuition tells me that `Err` value should be ignored like `Options`'s `None`.

MuhannadAlrusayni: I don't agree with this, the whole point on implementing `UpdateEl` to `Result` was to handle `Err` too.

if we didn't handle `Err` users might pass unhandled `Result` mistakenly and we simply have discard that `Err`.

MartinKavik: > if we didn't handle Err users might pass unhandled Result mistakenly and we simply have discard that Err.

Yes, I agree on 100%. 
But accidentally output a (long and cryptic) error to HTML which breaks layout and confuses user it even worse. It's pretty easy to do that - `String` and many other entities implements `UpdateEl`.
I begin to think that implementing `UpdateEl` for `Result` is really bad idea.

MuhannadAlrusayni: valid argument, I can delete the implementation for `Result`.
`Result` can be easily converted to `Option` (with `result.ok()`).

MartinKavik:
@MuhannadAlrusayni Look at https://github.com/seed-rs/seed/pull/373

MartinKavik:
@MuhannadAlrusayni Look at https://github.com/seed-rs/seed/pull/373

MartinKavik:
Closing in favor of https://github.com/seed-rs/seed/pull/373

rebo:
I think Option makes sense but I don't think it should be implemented for Result.

PR #384, opened 2020-03-09
Exclude use tag when routing

jgrund:
The `use` tag can contain an `href`. It should be excluded when looking
for the closest tag for routing purposes.

Signed-off-by: Joe Grund <jgrund@whamcloud.io>


MartinKavik:
Thanks! Just add a note into the changelog and I'll merge it.

PR #386, opened 2020-03-10
Updating the local storage interface

itsgreggreg:
* Does not expose panics
* Enables usage of both session and local storage
* Changes `name` to `key` in function parameters
* Adds a convenience methods for getting all keys, deleting a single key, and clearing all storage
* Adds more documentation

Hey guys this was something I ran into when writing an application that used local storage and I figured I'd take a stab at fixing it and cleaning up / extending the API a little. This does change the API so examples should probably be updated.

Anyway I figured I'd post what I've got here and let y'all comment or close this PR as you see fit. Happy to get it working up to the correct standards if you point me in the right direction!

### DIFF
--- a/src/browser/service/storage.rs
+++ b/src/browser/service/storage.rs
@@ -10,6 +10,7 @@ extern crate serde_json;
 
 pub type Storage = web_sys::Storage;
 
+#[deprecated(since = "0.7.0", note = "Use web_storage::get_local_storage")]
 #[allow(clippy::module_name_repetitions)]
 pub fn get_storage() -> Option<Storage> {
     web_sys::window()
@@ -20,6 +21,7 @@ pub fn get_storage() -> Option<Storage> {
 }
 
 /// Create a new store, from a serializable data structure.
+#[deprecated(since = "0.7.0", note = "use WebStorage.store_data")]
 pub fn store_data<T>(storage: &Storage, name: &str, data: &T)
 where
     T: serde::Serialize,
@@ -31,6 +33,7 @@ where
 }
 
 /// Load a store, to a deserializable data structure.
+#[deprecated(since = "0.7.0", note = "use WebStorage.load_data")]
 pub fn load_data<T>(storage: &Storage, name: &str) -> Option<T>
 where
     T: serde::de::DeserializeOwned,
--- a/examples/todomvc/src/lib.rs
+++ b/examples/todomvc/src/lib.rs
@@ -1,10 +1,6 @@
 use enclose::enc;
 use indexmap::IndexMap;
-use seed::{
-    browser::service::storage::{self, Storage},
-    prelude::*,
-    *,
-};
+use seed::{browser::service::web_storage::WebStorage, prelude::*, *};
 use serde::{Deserialize, Serialize};
 use std::mem;
 use uuid::Uuid;
@@ -37,7 +33,7 @@ struct Data {
 }
 
 struct Services {
-    local_storage: Storage,
+    local_storage: WebStorage,
 }
 
 #[derive(Default)]
@@ -95,8 +91,8 @@ fn init(url: Url, orders: &mut impl Orders<Msg>) -> Model {
         .subscribe(Msg::UrlChanged)
         .notify(subs::UrlChanged(url));
 
-    let local_storage = storage::get_storage().expect("get `LocalStorage`");
-    let data = storage::load_data(&local_storage, STORAGE_KEY).unwrap_or_default();
+    let local_storage = web_storage::get_local_storage().expect("get `LocalStorage`");
+    let data = local_storage.load(STORAGE_KEY).unwrap_or_default();
 
     Model {
         data,
@@ -209,7 +205,7 @@ fn update(msg: Msg, model: &mut Model, orders: &mut impl Orders<Msg>) {
         Msg::NoOp => (),
     }
     // Save data into LocalStorage. It should be optimized in a real-world application.
-    storage::store_data(&model.services.local_storage, STORAGE_KEY, &data);
+    model.services.local_storage.save(STORAGE_KEY, &data);
 }
 
 // ------ ------
--- a/src/browser/service.rs
+++ b/src/browser/service.rs
@@ -1,3 +1,4 @@
 pub mod fetch;
 pub mod routing;
 pub mod storage;
+pub mod web_storage;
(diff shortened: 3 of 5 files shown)


src/browser/service/storage.rs
```diff
@@ -12,33 +12,75 @@ pub type Storage = web_sys::Storage;
...
-pub fn store_data<T>(storage: &Storage, name: &str, data: &T)
+/// Clear all data in local storage
+pub fn clear(storage: &Storage) {
+    let _ = storage.clear();
+}
+
+/// All of the keys in local storage
+pub fn keys(storage: &Storage) -> Vec<String> {
+    let mut keys = vec![];
+    let length = storage.length().unwrap_or(0);
```
MartinKavik: does it make sense to continue when you can't get storage length? - Isn't `unwrap_or(0)` another silent fail? Let's fail fast. Or if it does make sense, we can use also `.unwrap_or_default()`.

src/browser/service/storage.rs
```diff
@@ -12,33 +12,75 @@ pub type Storage = web_sys::Storage;
...
 
-/// Create a new store, from a serializable data structure.
-pub fn store_data<T>(storage: &Storage, name: &str, data: &T)
+/// Clear all data in local storage
+pub fn clear(storage: &Storage) {
+    let _ = storage.clear();
+}
+
+/// All of the keys in local storage
+pub fn keys(storage: &Storage) -> Vec<String> {
```
MartinKavik: Does it make sense to return `Iterator` instead?

src/browser/service/storage.rs
```diff
@@ -12,33 +12,75 @@ pub type Storage = web_sys::Storage;
...
+    let mut keys = vec![];
+    let length = storage.length().unwrap_or(0);
+    for index in 0..length {
+        if let Ok(Some(key)) = storage.key(index) {
```
MartinKavik: Another silent fail?

src/browser/service/storage.rs
```diff
@@ -12,33 +12,75 @@ pub type Storage = web_sys::Storage;
...
+            keys.push(key);
+        }
+    }
+    keys
+}
+
+/// Things that can go wrong when trying to load data from local storage.
+pub enum LoadError {
+    /// Could not connect to local storage.
+    CouldNotConnect,
```
MartinKavik: What about `ConnectionError` to make it consistent with `DecodeError`?

src/browser/service/storage.rs
```diff
@@ -12,33 +12,75 @@ pub type Storage = web_sys::Storage;
...
+        Some(d) => {
+            let decoded = serde_json::from_str(&d);
+            decoded.map_err(LoadError::DecodeError)
+        }
+    }
+}
+
+/// Removes key and associated data from local storage
+pub fn remove_item(storage: &Storage, key: &str) {
+    let _ = storage.remove_item(key);
```
MartinKavik: Silent fail.

src/browser/service/storage.rs
```diff
@@ -12,33 +12,75 @@ pub type Storage = web_sys::Storage;
...
 }
 
-/// Load a store, to a deserializable data structure.
-pub fn load_data<T>(storage: &Storage, name: &str) -> Option<T>
+/// Things that can go wrong when trying to save data to local storage.
+pub enum SaveError {
+    /// The browser denied saving to local storage. Usually because the storage is full.
+    /// See: https://developer.mozilla.org/en-US/docs/Web/API/Storage/setItem#Exceptions
+    CouldNotSave,
```
MartinKavik: `SaveError` / `SetError`? Also we can save also exception content - `SetError(JsValue)` (applicable to all errors).

MartinKavik:
And please run `cargo make start verify` before push to format and lint your code. (It doesn't pass pipelines without it).

src/browser/service/storage.rs
```diff
@@ -12,33 +12,75 @@ pub type Storage = web_sys::Storage;
...
+pub fn remove_item(storage: &Storage, key: &str) {
+    let _ = storage.remove_item(key);
 }
 
-/// Load a store, to a deserializable data structure.
-pub fn load_data<T>(storage: &Storage, name: &str) -> Option<T>
+/// Things that can go wrong when trying to save data to local storage.
+pub enum SaveError {
+    /// The browser denied saving to local storage. Usually because the storage is full.
+    /// See: https://developer.mozilla.org/en-US/docs/Web/API/Storage/setItem#Exceptions
```
MartinKavik: Change it to something like `[MDN - Storage/setItem Exceptions](..` please

src/browser/service/storage.rs
```diff
@@ -12,33 +12,75 @@ pub type Storage = web_sys::Storage;
...
-        .expect("get `window`")
-        .local_storage()
-        .ok()
-        .flatten()
+    web_sys::window()?.local_storage().ok().flatten()
 }
 
-/// Create a new store, from a serializable data structure.
-pub fn store_data<T>(storage: &Storage, name: &str, data: &T)
+/// Clear all data in local storage
```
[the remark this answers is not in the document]
itsgreggreg: I'm 100% on the rename. Will do

itsgreggreg:
Thinking about this a bit more, it doesn't seem like there's much to offer on top of the existing `web_sys` implementation and perhaps simply deprecating the `storage` module in favor of directly using `web_sys` is the path forward here. Is there a guiding ux philosophy that seed has to follow to make this decision? Does seed want to wrap the implementation to ensure swapping out the jsbindings can happen without documentation changes?

MartinKavik:
I try to use `web_sys` directly as little as possible in Seed apps.
  - When I need to use `web_sys` I have to add it into app deps.
  - Then I have to enable required `web_sys` features (and there are MANY of them).
  - I have to choose compatible `web_sys` version with Seed and update it when needed.
  - It reflects weird Web API. 
  - `web_sys` functions often returns only `JsValue` - you have no idea what it represents until you read documentation and manually cast it into the correct type.
  - Once there is official support for native API access from WASM, you'll have to rewrite your code.
  - The main reason - Developer Experience - I think we can write better API with better error handling, better Seed integration, better autocomplete support, etc. In ideal world, developer don't have to read docs or copy-paste example, just explore API with autocomplete and doc comments and use it.

itsgreggreg:
Righteous, thanks for taking the time to explain that! All good reasons 👍🤙

MartinKavik:
@itsgreggreg Is it ready for code review?

MartinKavik:
Closing in favor of https://github.com/seed-rs/seed/pull/423.

PR #403, opened 2020-03-25
feat(fetch): Change `bearer` type to `Into<Cow>`

TatriX:
So user can use String and to mirror other header methods signatures.

### DIFF
--- a/src/browser/fetch/header.rs
+++ b/src/browser/fetch/header.rs
@@ -49,8 +49,8 @@ impl<'a> Header<'a> {
     }
 
     /// Create `Authorization: Bearer xxx` header.
-    pub fn bearer(token: &str) -> Header<'a> {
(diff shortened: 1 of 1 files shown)


MartinKavik:
Yeah, `&str` wasn't the best idea. I wanted to make it clear that `Cow` doesn't help you with the performance or anything else (it's just unnecessary step) because `format!` allocates a new `String`. `impl AsRef<str>` would be a better idea.

But I don't have any real problems with `Cow` - let me know what do you think about `AsRef<str>` - if you don't like it, I'll merge it.

TatriX:
Yep, I was thinking about `AsRef` as well, but decided to go with `Cow` just to make signatures look alike. As long as I can pass both `str` and `String` I'm fine with any implementation.

PR #418, opened 2020-04-13
Fix extraneous parentheses

ambiso:
I've noticed some redundant parentheses in some no_run annotated doctests, so I fixed them.

Feel free to close this issue if the changes are not needed.

Best,
ambiso


MartinKavik:
Thanks! Merging.

@ambiso Maybe we can remove `no_run` to force the compiler to check those comments / examples automatically - see https://doc.rust-lang.org/rustdoc/documentation-tests.html#hiding-portions-of-the-example. Do you think it would be something you would like to try?

PR #429, opened 2020-04-22
Fix bytes interface of Response

### LINKED ISSUE #427: Response::bytes throws `Error: expected a number argument`
Calling the `bytes()` function on the response of a fetch call, ie.
```rust
fetch(url).await.unwrap().bytes().await;
```
Causes the following error:
```
Stack:
Error: expected a number argument
    at _assertNum (http://localhost:8000/pkg/package.js:224:39)
    at imports.wbg.__wbg_length_b7dc6aed3ca09be1 (http://localhost:8000/pkg/package.js:1565:13)
    at js_sys::Uint8Array::length::h25f34bf3ef34d05a (wasm-function[3874]:0x17bcdf)
    at js_sys::Uint8Array::to_vec::hdf9eb12ced7141ca (wasm-function[1982]:0x13cfe3)
    at seed::browser::fetch::response::Response::bytes::{{closure}}::h188a81966303f00c (wasm-function[211]:0x6c06c)
```

I set a breakpoint and found that the `js_sys::Uint8Array` is actually wrapping an `ArrayBuffer`.
I believe the issue is that `Response::bytes` is calling `Uint8Array::from` instead of `Uint8Array::new` and it looks like js_sys then has an `ArrayBuffer` which it thinks is a `Uint8Array`. I'll make a PR and add a test.

platy:
Fixes #427 

Change from `Uint8Array::from` to `Uint8Array::new`.

I added a closure in order to borrow the ArrayBuffer, I don't know whether that is avoidable or desirable to avoid.

### DIFF
--- a/src/browser/fetch/response.rs
+++ b/src/browser/fetch/response.rs
@@ -50,7 +50,7 @@ impl Response {
             .map(JsFuture::from)?
             .await
             .map_err(FetchError::PromiseError)
-            .map(js_sys::Uint8Array::from)?
+            .map(|array_buffer| js_sys::Uint8Array::new(&array_buffer))?
             .to_vec())
     }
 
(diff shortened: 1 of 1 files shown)


src/browser/fetch/response.rs
```diff
@@ -96,3 +96,48 @@ impl Response {
...
+    use super::*;
+    use serde::Deserialize;
+
+    #[derive(Deserialize)]
+    struct Obj {
+        key: String,
+    }
+
+    #[wasm_bindgen_test]
+    async fn test_json() {
```
MartinKavik, line 113: Change test names please - e.g. `test_json` to `response_json`.
It will be more expressive while you are looking at the testing log and especially it allows you to run all response tests by prefix - e.g. `cargo make test_one_h response_`

MartinKavik:
> I added a closure in order to borrow the ArrayBuffer

Yeah, I hope Rust will be able to coerce passed value automatically in the future. So I think the closure is the best way now and it's also more expressive thanks to named argument.

I'll merge it once comment(s) are resolved, thanks!

platy:
Thanks for the review - resolved the comment.

PR #435, opened 2020-05-03
Libdoc

Ben-PH:
At the crate [documentation](https://docs.rs/seed/0.6.0/seed/), the top just says "See readme for details."

To be more ergonomic and helpful, I've added a link to the crate website, and the github (which also has the Readme), and brought in the Counter Example. 

No changes to the source code was made. 

It should be the case that the only changes outside documentation is with `cargo test --doc`. No changes to other doc-tests. new doc-test passes.


MartinKavik:
Thanks!

PR #447, opened 2020-05-09
Use correct case for XML (SVG) elements

klemens:
Previously XML elements which use some upper case letters would be converted to all-lowercase during string parsing. As xml is case-sensitive, this breaks the element. An example is the XML linearGradient.

This was caused by the code incorrectly converting all element names to lowercase to ensure HTML tags are correctly assigned, as they use lower case internally, but the browser function tagName returns all caps for HTML tags.

The first two commits fix this, while the last one changes the names of the gradient macros to camel case (like the other svg macros). That last one is also a breaking change.


MartinKavik:
Very nice! I just learned new stuff about Svg and Seed code 😄 
Please update changelog and I'll merge it.

PR #452, opened 2020-05-11
Proposal/RFC: Add more documentation, along with consolidating the module structure

Ben-PH:
## Suggested Changes

The libraries documentation leaves room to take advantage of the rusts doc tooling. This proposal shows two examples.

1. In-line documentation of modules at the crate root
2. Consolidating deep namespaces

#### In line documentation

At the root, it outlines top-level modules. These can make for for valuable 1 or 2 line comments. This can serve the valuable role of providing an overview of the APIs structure, and add context for how it's intended to be used.

#### Consolidation of namespaces

The list of re-exports at documentation don't add context to what's being exported. In this proposal, we add the `dom` module, provide context of what the module provides, and move re-exports into it. 

With this repeated for other re-exports, it will allow the front page to focus on providing context, with details intuitively organized into modules.

## Resolving Breaking Changes

There won't be any changes to the functionality of the library. Breaking changes would come from changes to the layout of the API. Internally, this is not of significant concern - the examples in this proposal were relatively trivial to put together, and don't break `cargo make verify`. Other changes were not so straight forward, though. 

Externally, the implications of these changes to the API structure are another matter. I'll leave it up to comments to propose how this could be managed

## Possible Implementation

The changes provided in this proposal PR were directly inspired by the way [rocket](https://rocket.rs/) handles its module and namespace structure. See the [code linked](https://github.com/SergioBenitez/Rocket/blob/1010f6a2a88fac899dec0cd2f642156908038a53/core/lib/src/lib.rs#L112) for an example.

## Next Steps

Immediately, there is nothing technical getting in the way of filling out the module inline documentation.

More ambitiously, consolidating the the module structure would be done along with adding to documentation. Given that this will impact the public API, this would need to be done carefully. There are other efforts underway to make changes to the API.

Integrating module consolidation into these plans, and discussion of what the final structure might be a good first start. I would prefer not to assume the needs of the project, though.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -89,16 +89,16 @@ macro_rules! map_callback_return_to_option_ms {
 // @TODO move to prelude (?)
 pub use crate::{
     app::{App, AppBuilder},
-    browser::dom::cast::{
-        to_drag_event, to_html_el, to_input, to_keyboard_event, to_mouse_event, to_select,
-        to_textarea, to_touch_event,
-    },
+    // browser::dom::cast::{
+    //     to_drag_event, to_html_el, to_input, to_keyboard_event, to_mouse_event, to_select,
+    //     to_textarea, to_touch_event,
+    // },
     browser::fetch,
     browser::service::routing::push_route,
     browser::service::storage,
     browser::url::Url,
     browser::util::{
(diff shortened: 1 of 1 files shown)


Ben-PH:
Closing as this is being actively addressed

PR #453, opened 2020-05-11
Tracking draft PR to remove deprecated functions

Ben-PH:
Added org file into this draft PR with notes, TODO's 

## Checklist 
TODO: edit this top-level comment to include things that need to be removed

- [x] CRITICAL: restore [clippys](https://github.com/seed-rs/seed/pull/453#discussion_r423548814)
- [x] `builder.rs` implementations cleared
- [x] `builder` module taken out of public API
- [x] `App::new()` in tests have init argument not Init code slated for deletion.
- [x] internal code `builder` dependencies removed
- [x] internal tests `builder` dependencies removed

- [x] `builder/after_mount.rs` implementations cleared
- [x] `builder/before_mount.rs` implementations cleared
- [x] `builder/init.rs` implementations cleared

tests pass running firefox. Chrome and safari unavailable.

Issue #277, opened 2019-11-03
Mount takeover behavior concerns.


AlterionX:
The code in question is introduced in PR #276.

A new enum `MountType` was introduced for dealing with how to handle elements that are already attached to the root element (the `mount_point`). The behavior of `MountType::Append` is approximately the same as before the change. The behavior of `MountType::Takeover` is that it will add preexisting nodes to the tree, delete the underlying DOM nodes, then construct new DOM nodes.

The second is a little problematic. It is problematic because there can be nodes that should not be recreated (like `script` nodes). It also replaces unknown elements with the `span` element, which may also be problematic if someone has custom tags.

Additionally, the tag checking mechanism is slightly buggy, as the many tags have an upper case letter in their spellings, causing scripts to be removed from the DOM. The tag names are listing in `dom_types.rs`.

Lastly, this is also a bit inefficient, but it is a one time cost.

These issues might need to be addressed later.

TatriX:
Here's a reproducible example based on `counter` example from the latest master:
```rust
// ... snip

fn before_mount(_: Url) -> BeforeMount {
    BeforeMount::new()
        .mount_point(seed::body())
        .mount_type(MountType::Takeover)
}


#[wasm_bindgen(start)]
pub fn render() {
    seed::log!("Started");
    App::builder(update, view).before_mount(before_mount).build_and_start();
}
```
You will see "Started" printed 2 times, while it should be printed only once.

MartinKavik:
Warning about scripts in the mount point: https://github.com/seed-rs/seed/pull/334/files

Ben-PH:
Does #478 close this issue?

PR #457, opened 2020-05-14
Delete builder module

Ben-PH:
Relates to #453

Concerns about #277 outlined in comments, and will need to be resolved before pulling in.


src/app.rs
```diff
@@ -176,73 +161,66 @@ impl<Ms, Mdl, INodes: IntoNodes<Ms> + 'static, GMs: 'static> App<Ms, Mdl, INodes
...
+        // Bootstrap the virtual DOM.
+        let mut orders = OrdersContainer::new(app.clone());
+        let model = init(Url::current().skip_base_path(&base_path), &mut orders);
+
+        app.data.model.replace(Some(model));
+        app.data
+            .main_el_vdom
+            .replace(Some(El::empty(Tag::Placeholder)));
+
+        // the ONLY place in the system this is checked.
```
Ben-PH, line 176: Missed this. Will amend.

src/app.rs
```diff
@@ -363,70 +336,6 @@ impl<Ms, Mdl, INodes: IntoNodes<Ms> + 'static, GMs: 'static> App<Ms, Mdl, INodes
         }
     }
 
-    /// Bootstrap the dom with the vdom by taking over all children of the mount point and
-    /// replacing them with the vdom if requested. Will otherwise ignore the original children of
-    /// the mount point.
-    fn bootstrap_vdom(&self, mount_type: MountType) -> El<Ms> {
```
Ben-PH, line 369: This is directly relevant to issue #277 

From the code that I was working with, `mount_type` was only ever passed in with `::Takeover` once `App::new()` was made private. Given that this case only ever returns the result of `El::empty(Tag::Placeholder)`, I opted to replace calls to this function with that.

src/app.rs
```diff
@@ -565,120 +474,4 @@ impl<Ms, Mdl, INodes: IntoNodes<Ms> + 'static, GMs: 'static> App<Ms, Mdl, INodes
...
-        Self::builder(update, view).init(Box::new(init))
-    }
-
-    /// App initialization: Collect its fundamental components, setup, and perform
-    /// an initial render.
-    #[deprecated(
-        since = "0.4.2",
-        note = "Please use `AppBuilder.build_and_start` instead"
-    )]
-    pub fn run(mut self) -> Self {
```
Ben-PH, line 587: This was moved up into `App::start`. Future efforts will want to simplify this, perhaps?

src/app/cfg.rs
```diff
@@ -1,19 +1,16 @@
 use super::types::{SinkFn, UpdateFn, ViewFn, WindowEventsFn};
-use super::{builder::IntoAfterMount, MountType};
 use crate::virtual_dom::IntoNodes;
 use std::marker::PhantomData;
 use std::rc::Rc;
 
 #[allow(clippy::module_name_repetitions)]
-pub struct AppInitCfg<Ms, Mdl, INodes, GMs, IAM: ?Sized>
```
Ben-PH, line 8: Removed along with `IntoAfterMount`

src/app/orders/container.rs
```diff
@@ -23,11 +23,6 @@ impl<Ms, Mdl, INodes: IntoNodes<Ms>, GMs> OrdersContainer<Ms, Mdl, INodes, GMs>
             app,
         }
     }
-
-    pub(crate) fn merge(&mut self, mut other: Self) {
```
Ben-PH, line 27: got picked up as "never called"

src/app.rs
Ben-PH: ammended

src/app.rs
```diff
@@ -176,73 +161,64 @@ impl<Ms, Mdl, INodes: IntoNodes<Ms> + 'static, GMs: 'static> App<Ms, Mdl, INodes
...
-            update,
-            None,
-            view,
-            root_element,
-            None,
-            None,
-            Some(app_init_cfg),
-            base_path,
+        let root_element = root_element.get_element().expect("get root element");
+        let app = Self::new(update, view, root_element, base_path.clone());
```
MartinKavik, line 165: Please use `Rc::clone` where possible (see the old code in this diff).

https://doc.rust-lang.org/book/ch15-04-rc.html#using-rct-to-share-data:
> We could have called a.clone() rather than Rc::clone(&a), but Rust’s convention is to use Rc::clone in this case. The implementation of Rc::clone doesn’t make a deep copy of all the data like most types’ implementations of clone do. The call to Rc::clone only increments the reference count, which doesn’t take much time. Deep copies of data can take a lot of time. By using Rc::clone for reference counting, we can visually distinguish between the deep-copy kinds of clones and the kinds of clones that increase the reference count. When looking for performance problems in the code, we only need to consider the deep-copy clones and can disregard calls to Rc::clone.

src/app.rs
```diff
@@ -324,26 +300,21 @@ impl<Ms, Mdl, INodes: IntoNodes<Ms> + 'static, GMs: 'static> App<Ms, Mdl, INodes
...
         let document = window.document().expect("get window's document");
 
         Self {
-            init_cfg,
             cfg: Rc::new(AppCfg {
                 document,
                 mount_point,
                 update,
-                sink,
+                sink: None,
```
MartinKavik, line 315: Just note to not forget remove `sink`, `window_events` and `routes`.

src/app.rs
```diff
@@ -1,27 +1,20 @@
...
+use enclose::enclose;
 use std::{
     any::Any,
     cell::{Cell, RefCell},
     collections::VecDeque,
-    marker::PhantomData,
     rc::Rc,
 };
-use types::{RoutesFn, SinkFn, UpdateFn, ViewFn, WindowEventsFn};
+use types::{UpdateFn, ViewFn};
```
MartinKavik, line 14: We should consider to refactor / remove `types` module.

src/app.rs
```diff
@@ -144,12 +127,16 @@ impl<Ms, Mdl, INodes: IntoNodes<Ms> + 'static, GMs: 'static> App<Ms, Mdl, INodes
...
     ///
+    #[allow(clippy::too_many_lines)]
     pub fn start(
         root_element: impl GetElement,
         init: impl FnOnce(Url, &mut OrdersContainer<Ms, Mdl, INodes, GMs>) -> Mdl + 'static,
         update: UpdateFn<Ms, Mdl, INodes, GMs>,
         view: ViewFn<Mdl, INodes>,
     ) -> Self {
+        // This function looks to be a significant sticking point. It will possibly end
+        // up being a relatively major effort to get it cleaned up.
```
MartinKavik, line 138: I agree, we can do it in the next PR.

src/app.rs
Ben-PH: next PR, or add it to this one?

MartinKavik: I don't know how hard it would be - your choice.

src/app.rs
Ben-PH: base_path is already an Rc. Are you talking about the app itself?

That sould work for the app, but I think it might be digging ourselves into a hole. At the moment, making a new app is dependent on having a model, but once you have that app, it needs to take ownership of the model. `Rc` can help us get around it, but it feels a bit hacky - kinda like a circular dependency. I suspect that there's a better way to do it.

Outside of this piece of code, the nature of the data doesn't seem to create any need of a ref-count. I think part of cleaning up `start` should include clearing away deep-copies and unnecisary `Rc` wrapping. 

How about we pull this into a development branch? That way we can have a go at not having deep-copy AND not wrapping it in `Rc`. If it doesn't work out, we can fix it up before bringing it into production.

src/app.rs
Ben-PH: So I sorted that out. sink and window_events i don't recall making too much difference (my memory could be wrong there, so take with a grain of salt). Culling `routes`, though, cascaded into a few other code changes that come with broader consequences.

src/app.rs
```diff
@@ -299,18 +267,6 @@ impl<Ms, Mdl, INodes: IntoNodes<Ms> + 'static, GMs: 'static> App<Ms, Mdl, INodes
         }
     }
 
-    pub fn patch_window_event_handlers(&self) {
```
Ben-PH, line 302: Remember: we drop `window_events`, making this redundant.

src/app.rs
```diff
@@ -287,8 +255,8 @@ impl<Ms, Mdl, INodes: IntoNodes<Ms> + 'static, GMs: 'static> App<Ms, Mdl, INodes
                     let mut new_effects = self.process_queue_message(msg);
                     queue.append(&mut new_effects);
                 }
-                Effect::GMsg(g_msg) => {
```
Ben-PH, line 290: I don't understand the broader system well enough to recognise what this change implies. A grep of `Effect::GMsg` came up with 4 hits:

1. This one
2. Pushing a message onto the global message queue [here](https://github.com/Ben-PH/seed/blob/30614b1d24356c2d2225e6f4cf526d9405ac3121/src/app/orders/container.rs#L102)
3. [10 lines up](https://github.com/Ben-PH/seed/blob/30614b1d24356c2d2225e6f4cf526d9405ac3121/src/app.rs#L247) pushing onto front as part of the `sink` function, which I understand we plan to remove.
4. does [nothing](https://github.com/seed-rs/seed/blob/master/src/app/effects.rs#L20)

So, in terms of utility, summarizing the use of global message

1. We are discarding it as an unused variable
2. We are storing the value into a queue.
3. As far as I can tell, this is the only place that the value is used. Everywhere else, it's private and seems to only come after calling `fn sink`
4. is effectively a no-op as far as GM is concerned

#### Interesting...
Aside from running `self.data.msg_listeners` callbacks, and running the `update` function, the code is the same, but not if we consider `OrdersContainer` construction because it always gives `ShouldRender::Render`. This means that all it does is call `self.schedule_render()` and return an empty queue. 

#### Possible issues?

Well, maybe. I can't see how the system uses the value that `::GMsg` wraps, and I feel a bit blind as to the motivations to put in `sink`in the first place. What are we throwing away here?

src/app.rs
Ben-PH: I'll leave it to another PR. Taking out `window_events` might reveal some bugs, and I we _might_ be biting off more than we can chew already. 

I'm happy for this PR to be put onto a dev-branch, then bring in a bunch of tests for the examples (and real-world app, I think) to make sure everything's okay. Then we can look at types. How does that sound?

src/app.rs
MartinKavik: > base_path is already an Rc. Are you talking about the app itself?

Please read my comment and copy-pasted explanation again :)

`Rc::clone(&base_path)` vs `base_path.clone()` => just replace `.clone` with `Rc::clone`.

src/app.rs
MartinKavik: `sink`, `Effect::GMsg`, `<...GMs..>`, `send_g_gms`, `perform_g_cmd`and `perform_g_cmd_with_handle` should be removed.

Global messages work the same like regular messages (`Msg` / `Ms`), but the whole app use only one type. This mechanism of inter-component communication has been replaced by `orders.subscribe / notify`. Everything  `GMsg`-related will be deleted.

src/app.rs
MartinKavik: Try `cargo make start subscribe` => http://localhost:8000/ => resize window => check that `Window size: ww x hh` has been changed.

src/app.rs
MartinKavik: I think this PR should be merged into `master` before we'll create a new one. You don't want to resolve complicated conflicts once somebody push bigger changes to `master` and forgot context while you are working on other parts in a new PR.

Ben-PH: Hmmm, I can only disagree from a place of relative lack of experience. I can see the logic in it.

Ben-PH:
Closing, as breakages have come up.

PR #467, opened 2020-05-25
Examples/charts

TatriX:
Here's a chart example.
It's a bit messy right now, as a lot of things are hardcoded and not  properly structured.
But it should be good enough to serve as a starting point.

2020-05-25-145127_3077x1267_scrot
As you can see in the dev tools, there is a `xmlns` attribute on every svg node. I believe all of those can be removed: see "note" on the [mdn](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/svg)

### DIFF
--- a/examples/charts/src/lib.rs
+++ b/examples/charts/src/lib.rs
@@ -8,15 +8,15 @@ type Point = (f64, f64);
 
 struct Model {
     line_data: Vec<Point>,
-    tooltip: Option<(i32, i32)>,
+    tooltip: Option<line::Tooltip>,
 }
 
 impl Default for Model {
     fn default() -> Self {
-        let end = 6.0 * std::f64::consts::PI;
+        let end = 2.0 * std::f64::consts::PI;
         let n = 300;
         Self {
-            line_data: (0..n)
+            line_data: (0..=n)
                 .map(|i| {
                     let x = f64::from(i) * end / f64::from(n);
                     let y = f1(x);
@@ -29,20 +29,20 @@ impl Default for Model {
 }
 
 fn f1(x: f64) -> f64 {
-    x.sin()
+    x.sin() + 1.0
 }
 
 #[derive(Clone)]
 enum Msg {
-    ShowTooltip(i32, i32),
+    ShowTooltip(line::Tooltip),
     HideTooltip,
 }
 
 fn update(msg: Msg, model: &mut Model, _: &mut impl Orders<Msg>) {
     match msg {
-        Msg::ShowTooltip(x, y) => {
-            model.tooltip = Some((x, y));
-        },
+        Msg::ShowTooltip(tooltip) => {
+            model.tooltip = Some(tooltip);
+        }
         Msg::HideTooltip => {
             model.tooltip = None;
         }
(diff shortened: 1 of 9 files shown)


examples/charts/src/lib.rs
```diff
@@ -0,0 +1,82 @@
...
+        }
+        Msg::HideTooltip => {
+            model.tooltip = None;
+        }
+    }
+}
+
+fn view(model: &Model) -> Node<Msg> {
+    div![
+        style! {St::Width => "80vh", St::Height => "50vh", St::Margin => "auto"},
```
MartinKavik, line 54:
```suggestion
        style! {St::Width => vh(80), St::Height => vh(50), St::Margin => "auto"},
```

examples/charts/src/line.rs
```diff
@@ -0,0 +1,140 @@
+use itertools::Itertools;
+use seed::{prelude::*, *};
+
+#[derive(Clone)]
```
MartinKavik, line 4:
```suggestion
#[derive(Copy, Clone)]
```

examples/charts/src/line.rs
```diff
@@ -0,0 +1,140 @@
...
+    pub position: (i32, i32),
+    pub data: (f64, f64),
+}
+
+#[allow(clippy::too_many_lines)]
+pub fn chart<T: Clone + 'static>(
+    data: &[(f64, f64)],
+    onenter: impl Fn(Tooltip) -> T + Clone + 'static,
+    onout: T,
+) -> Node<T> {
```
MartinKavik, line 15:
```suggestion
pub fn chart<Ms: Clone + 'static>(
    data: &[(f64, f64)],
    on_enter: impl Fn(Tooltip) -> Ms + Clone + 'static,
    on_out: Ms,
) -> Node<Ms> {
```

examples/charts/src/line.rs
```diff
@@ -0,0 +1,140 @@
...
+                        circle![attrs! {
+                            Cx => x / dx + f64::from(padding),
+                            Cy => -y / dx + f64::from(height - padding),
+                            R => 2,
+                            Fill => "rgba(0, 86, 91, 0.8)",
+                        }],
+                    ])
+                } else {
+                    None
+                })
```
MartinKavik, line 137:
```suggestion
                .filter_map(|(i, (x, y))| IF!(i % 15 == 0 => {
                    let onenter = onenter.clone();
                    let onout = onout.clone();
                    let data = (*x, *y);
                    Some(g![
                        C!["data-point"],
                        mouse_ev(Ev::MouseOver, move |event| {
                            onenter(Tooltip {
                                position: (event.x(), event.y()),
                                data,
                            })
                        }),
                        mouse_ev(Ev::MouseOut, move |_| onout),
                        circle![attrs! {
                            Cx => x / dx + f64::from(padding),
                            Cy => -y / dx + f64::from(height - padding),
                            R => 2,
                            Fill => "rgba(0, 86, 91, 0.8)",
                        }],
                    ])
                }))
```

examples/charts/src/line.rs
```diff
@@ -0,0 +1,138 @@
...
+                        -y / dx + f64::from(height - padding),
+                    ))
+                    .collect::<Vec<_>>()
+                    .join(" ")
+            }],
+            data.iter()
+                .enumerate()
+                .filter_map(|(i, (x, y))| IF!(i % 15 == 0 => {
+                    let onenter = onenter.clone();
+                    let onout = onout.clone();
```
MartinKavik, line 117: on_enter, on_out

flosse:
closing in favor of #664

PR #468, opened 2020-05-25
Add example i18n leveraging fluent to implement multilingual support.

tokcum:
As discussed in forum [Multilingual support](https://seed.discourse.group/t/multilingual-support/78/8).

I'm aware of CONTRIBUTING.md. Unfortunately the fork from master branch didn't compile with "cargo make verify" on my machine before adding the i18n example.


examples/i18n/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+
+mod resource;
+
+// ------ ------
+//     Init
+// ------ ------
+
+fn init(_: Url, _: &mut impl Orders<Msg>) -> Model {
+    let mut model = Model::default();
+    change_lang(&mut model);
```
MartinKavik, line 20: Rust supports safe mutability but use it only when best practices for immutable functions and testability make the code too unreadable or slow.
I would rather see something like:

```rust
const DEFAULT_LANG: Lang = Lang::en_US;
...
fn init(_: Url, _: &mut impl Orders<Msg>) -> Model {
    let i18n = I18n::new(DEFAULT_LANG);
    Model { i18n }
}
```
or 
```rust
fn init(_: Url, _: &mut impl Orders<Msg>) -> Model {
    Model { 
        i18n: I18n::new(DEFAULT_LANG),
    }
}
```
where
```rust

struct Model {
    i18n: I18n
}

// ------ I18n ------

struct I18n {
    lang: Lang,
    resource: ...
}
```
=> No more implicit (and error-prone) behavior and you can't forget to call `change_lang`. Also writing unit tests for `change_lang` would be pain once your `Model` is big enough.

examples/i18n/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+version = "0.1.0"
+authors = ["Tobias Mucke <tobias.mucke@gmail.com>"]
+edition = "2018"
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+seed = { path = "../../" }
+enum-iterator = "~0.6"
```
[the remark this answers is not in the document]
MartinKavik: Ad `enum-iterator` - once it's not enough for you, I recommend to use https://crates.io/crates/strum

examples/i18n/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+    de_DE,
+}
+impl Lang {
+    fn id(&self) -> &str {
+        match self {
+            Lang::en_US => "en-US",
+            Lang::de_DE => "de-DE",
+        }
+    }
+    fn label(&self) -> String {
```
MartinKavik, line 40: Why `String` instead of `&str`?

examples/i18n/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+        }
+    }
+}
+
+pub struct Model {
+    lang: Lang,
+    resource: FluentBundle<FluentResource>,
+}
+
+impl Default for Model {
```
MartinKavik, line 53: Remove `Default`. See review comments for `init` function above.

examples/i18n/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+    let mut model = Model::default();
+    change_lang(&mut model);
+    model
+}
+
+// ------ ------
+//     Model
+// ------ ------
+#[allow(non_camel_case_types)]
+#[derive(Clone, IntoEnumIterator, PartialEq)]
```
MartinKavik, line 28: Derive also `Copy`.

examples/i18n/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+enum Lang {
```
MartinKavik, line 29: Please move `Lang` below `Model` and add comment `// ------ Lang ------` above it (you can see it in other examples).

You want to see the "parent" item first (`Model` in this case) to see usage (`lang: Lang`) and then you want to see the definition (`enum Lang...`). It's not super clear when there is only one "child", but you can't load all children into your short-term memory when there are many child items. It would cause increased coginitive load and jumping from one place in code to another and back.
The same rule is valid for all nested items like nested `view`s and `update` helpers.

MartinKavik:
OT: @tokcum Your commit has been pushed from different account (probably your GitHub account and git client has different email addresses or something like that). I recommend to fix it so you get all credits for the contributions and the commit history and PRs would be a little bit cleaner.

tokcum:
@MartinKavik thank you for your advice and feedback. I synced the settings of my Git client with my GitHub account. I'm going to work on your comments to resolve the issues.

examples/i18n/Cargo.toml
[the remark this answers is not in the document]
tokcum: Check, switched to fully nailed down version strings and migrated from enum-iterator to strum.

examples/i18n/README.md
```diff
@@ -0,0 +1,11 @@
+## I18N example
+
+Intended as a demo of basic functionality.
```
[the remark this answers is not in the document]
tokcum: Sure. README is now more expressive and global README is adopted.

examples/i18n/src/lib.rs
tokcum: Sure. Done.

examples/i18n/src/lib.rs
tokcum: Good reasoning. Done.

examples/i18n/src/lib.rs
tokcum: Legitimate question. No answer. ;) Done.

examples/i18n/src/lib.rs
tokcum: Good lesson. I got rid of stand alone function "change_lang" by implementing a function to set lang within I18n. Still I struggle with I18n::new(). Please check again my implementation. Thanks.

examples/i18n/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+use std::borrow::Borrow;
+
+mod resource;
+
+// ------ ------
+//     Init
+// ------ ------
+
+fn init(_: Url, _: &mut impl Orders<Msg>) -> Model {
+    let mut model = Model::default();
```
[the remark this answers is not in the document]
tokcum: Agree. Done.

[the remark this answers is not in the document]
tokcum: You could post the details in a raw form here if you like. I could do some research on it, weighing up the pros and cons and write a nice post for seed-rs.org. I guess this is another PR because seed-rs.org is itself on GitHub? Just an offer, no offence.

tokcum:
> OT: @tokcum Your commit has been pushed from different account (probably your GitHub account and git client have different email addresses or something like that). I recommend to fix it so you'll get all credits for the contributions and the commit history and PRs would be a little bit cleaner.

Where have you seen this mismatch? Is it okay now?

examples/i18n/src/lib.rs
tokcum: Done.

examples/i18n/src/lib.rs
```diff
@@ -0,0 +1,179 @@
+// Some Clippy linter rules are ignored for the sake of simplicity.
+#![allow(clippy::needless_pass_by_value, clippy::trivially_copy_pass_by_ref)]
```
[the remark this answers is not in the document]
tokcum: This line was just copied over from counter example. Deleted. I ran "cargo make verify" directly after cloning seed repo. It fails with 21 errors mostly related to thread safety.

Not sure what is going on. Maybe there is something wrong in my dev environment?

`
[cargo-make] INFO - Execute Command: "cargo" "clippy" "--all-features" "--" "--deny" "warnings" "--deny" "clippy::pedantic" "--deny" "clippy::nursery"
    Checking seed v0.7.0 (/home/tobias/IdeaProjects/seed)
error: future cannot be sent between threads safely
  --> src/app/cmds.rs:26:6
   |
26 | ) -> impl Future<Output = MsU> {
   |      ^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `-D clippy::future-not-send` implied by `-D clippy::nursery`
   = note: `impl FnOnce() -> MsU + Clone + 'static` doesn't implement `std::marker::Send`
   = note: `*mut u8` doesn't implement `std::marker::Send`
   = note: `(dyn std::ops::FnMut() + 'static)` doesn't implement `std::marker::Send`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#future_not_send

error: this could be a `const fn`
   --> src/browser/fetch/request.rs:159:5
    |
159 | /     pub fn timeout(mut self, timeout: u32) -> Self {
160 | |         self.timeout = Some(timeout);
161 | |         self
162 | |     }
    | |_____^
    |
    = note: `-D clippy::missing-const-for-fn` implied by `-D clippy::nursery`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#missing_const_for_fn

[...]

error: aborting due to 21 previous errors

error: could not compile `seed`.

To learn more, run the command again with --verbose.
[cargo-make] ERROR - Error while executing command, exit code: 101
[cargo-make] WARN - Build Failed.
`

examples/i18n/src/resource/mod.rs
```diff
@@ -0,0 +1,66 @@
+use std::collections::HashMap;
+
```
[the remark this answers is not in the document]
tokcum: Ok, I refactored the example.

tokcum:
@MartinKavik , thanks for your feedback. I marked all comments as resolved in case it was clear to me how to get things done. Two comments are still open despite I worked on your feedback

1. What is the issue related to "cargo make verify"?
2. What are the pros / cons of default() over new() in Seed context?

I'm sure you have even more comments on my code. Ready, vive me more.

MartinKavik:
Nice, I'll look at it tommorow, it's midnight here 👌🌃

examples/i18n/src/lib.rs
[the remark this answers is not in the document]
MartinKavik: According to Clippy [changelog](https://github.com/rust-lang/rust-clippy/blob/master/CHANGELOG.md), the lint rule `future_not_send` is a new one and available only on Clippy beta / Rust 1.44. So I think you use nightly Rust, switch to stable and it should be ok. 
And I don't want to resolve these new Clippy problems because it may cause conflicts with the stable Clippy.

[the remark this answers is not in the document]
MartinKavik: I'm not sure if there is explicitly said in the CONTRIBUTING.md that you should switch to `stable` Rust - if it's missing I would be glad for PR.

examples/i18n/src/lib.rs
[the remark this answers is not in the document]
MartinKavik: I though I've already written it but not yet, so I'll copy/paste it probably from here into seed-rs.org PR :)
So..

---

Generally all implementations of standard traits (`From`, `Into`, `Default`, `Display`) are very useful if the item (`struct`, `enum`...) is used in multiple contexts or with multiple other items - then the generalization makes sense because it implies you are writing idiomatic Rust and it plays nicely with other standard traits and other items.
However when you start to implement standard traits for many items, your code-base is slowly turning into the sea of `.into()`, `::default()`, `.to_string()`, etc. 
As the result:

   - You lose expressive domain-specific names so it can be pretty hard to orient in the code.
   - It bloats the code because standard traits have to cover many cases so they tend to be more complicated.
   - You need to "bend" some parts of your code so it can be written with those implemented traits - it also makes the code harder to read any probably slower.

Ad the `Default` trait: We assume `xx::default()` calls are pretty cheap operations (see `Default` for [primitive types](https://doc.rust-lang.org/src/core/default.rs.html#132) or [Vec](https://doc.rust-lang.org/src/alloc/vec.rs.html#2334-2339)) - in the most cases there isn't even memory allocation on the heap and you probably won't find more expensive operations in `Default` implementations for other items. So when you write more sophisticated `Default` code for your item and somebody use this item in a nested structure, he will be very surprised once he writes some benchmarks.
One Seed user was even able to accidentally write recursive loop of nested complex `Default`s that was causing stack overflow.
In Seed apps, you need to create `Model` only once, so when you implement `Default` for `Model`:
  - It only bloats the code.
  - You'll have tendency to overwrite default values in `init` function because some `Model` parts will depend on `Url` or on other values => worse readability and slower code.
  - It sends misleading signal that `Model` is/can be created on multiple places.

---

> write a nice post for seed-rs.org

I'll announce the draft in #general chat channel once ready - then I'll probably merge it into seed-rs.org `master` and you'll be able to fix it / extend it through classic PRs. And then (probably with Seed 0.8.0 release) we want to focus also on marketing more - if you like to write, help would be very welcome.

examples/i18n/src/resource/german.rs
```diff
@@ -0,0 +1,44 @@
+pub static S: &'static str = r#"
```
MartinKavik, line 1: 1. Remove `pub static S: &'static str = r#"` and `"#`
1. Rename the file to `german.ftl`
1. Perhaps create a new folder `resources` or `translations` in the root or in `i18n` folder and move this file into it.
1. Replace `(german::S).to_string()` in `resource.rs` with `include_str!("path_to_the_file").to_string()`

examples/i18n/src/resource.rs
```diff
@@ -0,0 +1,66 @@
+use std::collections::HashMap;
+
+mod german;
+
+#[derive(Debug)]
+pub struct Resource(HashMap<String, String>);
+
+impl Default for Resource {
+    fn default() -> Self {
+        let ftl = String::from(
```
MartinKavik, line 10: I assume the FTL below are English translations => extract it to the file `english.ftl` (see review comment about `german.ftl` below)

examples/i18n/README.md
```diff
@@ -0,0 +1,17 @@
+## I18N example
+
+How to support multiple languages in your web app based on [Fluent][url_project_fluent]. 
+Includes a language selector, some sample text and FTL strings demonstrating
```
MartinKavik, line 4: Please attach a link to `FTL string` that points to docs - I have no idea what's that until I learn `fluent`. (In both READMEs)

examples/i18n/README.md
```diff
@@ -0,0 +1,17 @@
+## I18N example
+
+How to support multiple languages in your web app based on [Fluent][url_project_fluent]. 
+Includes a language selector, some sample text and FTL strings demonstrating 
+the simplicity and power of [Seed][url_project_seed] powered by [Fluent's rust crate][url_crate_fluent].
```
MartinKavik, line 5: I would write `Rust` or remove `rust`.

examples/i18n/src/lib.rs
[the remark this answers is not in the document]
tokcum: Ah, that's it. I'm using nightly Rust. Agree. I send a different PR for CONTRIBUTING.md.

examples/i18n/src/lib.rs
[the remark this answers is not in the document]
tokcum: Looking forward to contribute to it. I started with Seed 0.6 and at that time had only little experience with Rust in general and Web Assembly particularly.
I like Seed's mission statement "Complete documentation that always matches the current version. Getting examples working, and starting a project should be painless, and require nothing beyond this guide."
Today some of the documentation found on seed-rs.org still reflects 0.6 despite 0.7 release is available. It would be great to bridge the gap to the mission statement.

I've seen some crates which use pre releases to be able to get everything in sync with the release. Maybe this could be an organisational measure to facilitate release management.

examples/i18n/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+impl I18n {
+    fn new(lang: Lang) -> Self {
+        let mut i18n = I18n {
+            lang,
+            resource: FluentBundle::default(),
+        };
+        i18n.lang(lang);
+        i18n
+    }
+    fn lang(&mut self, lang: Lang) -> &Self {
```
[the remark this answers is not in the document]
tokcum: Found and done. I'll also add it as a resource to CONTRIBUTING.md in dedicated PR mentioned above.

examples/i18n/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+
+        self.resource = bundle;
+        self
+    }
+}
+
+// ------ Lang ------
+
+#[allow(non_camel_case_types)]
+#[derive(Clone, Copy, EnumIter, PartialEq)]
```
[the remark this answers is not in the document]
tokcum: I was not aware that derive can be order-sensitive. Done.

examples/i18n/src/lib.rs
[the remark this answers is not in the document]
tokcum: Done.

examples/i18n/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+        self.lang = lang;
+        let res = FluentResource::try_new(
+            resource::Resource::new()
+                .get(lang.id().to_string().borrow())
+                .unwrap()
+                .parse()
+                .unwrap(),
+        )
+        .expect("Failed to parse an FTL string.");
+        let locale: LanguageIdentifier = lang.id().parse().expect("Parsing failed");
```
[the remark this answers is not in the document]
tokcum: Done.

examples/i18n/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+    args_female_pl.insert("tabCount", FluentValue::from(7));
+    args_female_pl.insert("formal", FluentValue::from("false"));
+
+    div![
+        div![select![
+            attrs! {At::Name => "lang"},
+            langs,
+            input_ev(Ev::Change, Msg::LangChanged),
+        ],],
+        div![p!["Language in Model: ", model.i18n.lang.label()]],
```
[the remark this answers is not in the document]
tokcum: Ok. I implemented a getter method lang() to set lang. I also refactored the structure and created another module for I18n to make fields private and control usage via public methods. I hope this didn't move into wrong direction.

examples/i18n/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+        .get_message(label)
+        .expect("Message doesn't exist.");
+    let mut errors = vec![];
+    let pattern = msg.value.expect("Message has no value.");
+
+    let value = model
+        .i18n
+        .resource
+        .format_pattern(&pattern, args, &mut errors);
+    value.into()
```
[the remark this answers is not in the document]
tokcum: Done.

examples/i18n/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+                "tabs-close-warning"
+            )],
+            p![translate(model, None, "sync-dialog-title")],
+            p![translate(model, None, "sync-headline-title")],
+            p![translate(model, None, "sync-signedout-title")],
+        ],
+    ]
+}
+
+fn translate(model: &Model, args: Option<&FluentArgs>, label: &str) -> String {
```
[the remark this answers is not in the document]
tokcum: Agree. Nice practice to simplify interface and facilitate separation of modules. Done.

examples/i18n/README.md
tokcum: I like to mention Rust. :) Done.

examples/i18n/README.md
tokcum: Sure. Link to FTL syntax guide added.

examples/i18n/src/resource/german.rs
tokcum: Wow. I didn't expect this to be so easy. Works like a charm.

examples/i18n/src/resource.rs
tokcum: Yes! Done.

examples/i18n/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+    args_male_sg.insert("userGender", FluentValue::from("male"));
+    args_male_sg.insert("tabCount", FluentValue::from(1));
+    args_male_sg.insert("formal", FluentValue::from("true"));
+
+    let mut args_female_pl = FluentArgs::new();
+    args_female_pl.insert("userName", FluentValue::from("Anna"));
+    args_female_pl.insert("photoCount", FluentValue::from(5));
+    args_female_pl.insert("userGender", FluentValue::from("female"));
+    args_female_pl.insert("tabCount", FluentValue::from(7));
+    args_female_pl.insert("formal", FluentValue::from("false"));
```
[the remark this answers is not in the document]
tokcum: Thanks. I didn't stumble over this macro while skimming through the fluent docs. Done.

tokcum:
@MartinKavik this round goes to you. I was not able to work on all your comments today. Many thanks for your support. I guess you would have written this example in less time compared to coaching my Rust journey. I'm keen to learn the Rust way. And this PR is part of it.

examples/i18n/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+                .unwrap(),
+        )
+        .expect("Failed to parse an FTL string.");
+        let locale: LanguageIdentifier = lang.id().parse().expect("Parsing failed");
+        let mut bundle = FluentBundle::new(&[locale]);
+        bundle
+            .add_resource(res)
+            .expect("Failed to add FTL resources to the bundle.");
+
+        self.resource = bundle;
```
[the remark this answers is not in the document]
tokcum: These were different kind of resources. I clarified the code by using more specific var names. I don't see how to shorten this part of the code. Maybe there is a more rusty way?

examples/i18n/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+        };
+        i18n.lang(lang);
+        i18n
+    }
+    fn lang(&mut self, lang: Lang) -> &Self {
+        self.lang = lang;
+        let res = FluentResource::try_new(
+            resource::Resource::new()
+                .get(lang.id().to_string().borrow())
+                .unwrap()
```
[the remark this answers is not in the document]
tokcum: Done.

examples/i18n/src/lib.rs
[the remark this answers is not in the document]
tokcum: Thank you for the explanation regarding these default traits and why they are there on the one hand and should be used only in specific cases on the other hand. I have read a lot of Rust docs online and even on paper. A lot of things were explained in great detail but I missed so sadly such background knowledge or lets call it Dos and Don'ts. Maybe I have overlooked it. I'll check again in the next days.

examples/i18n/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+                .parse()
+                .unwrap(),
+        )
+        .expect("Failed to parse an FTL string.");
+        let locale: LanguageIdentifier = lang.id().parse().expect("Parsing failed");
+        let mut bundle = FluentBundle::new(&[locale]);
+        bundle
+            .add_resource(res)
+            .expect("Failed to add FTL resources to the bundle.");
```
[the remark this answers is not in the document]
tokcum: Sure. I shortened all expect messages.

examples/i18n/src/lib.rs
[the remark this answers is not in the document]
MartinKavik: > Today some of the documentation found on seed-rs.org still reflects 0.6 despite 0.7 release is available. It would be great to bridge the gap to the mission statement.

It's basically a unique situation because it's the first time I'm writing docs on seed-rs.org - and I decided to rewrite it from scratch.  Also there wasn't a way how to write docs for multiple versions until I added version switcher a few days / weeks ago. So this shouldn't happen again.

> I've seen some crates which use pre releases to be able to get everything in sync with the release.

Yeah I was thinking about something like that, once release problems appear again we can try it. Thanks for the tip!

> A lot of things were explained in great detail but I missed so sadly such background knowledge or lets call it Dos and Don'ts. Maybe I have overlooked it. I'll check again in the next days.

I also didn't find anything like that. That's why I've written it here from my experience from working on Seed and freelance projects and you'll find many similar cases in new seed-rs.org docs once ready. So don't kill time with searching Dos/Don'ts, I hope you'll find them on seed-rs.org (I hope in weeks max) and give me some feedback :)

examples/i18n/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+    LangChanged(String),
+}
+
+fn update(msg: Msg, model: &mut Model, _: &mut impl Orders<Msg>) {
+    match msg {
+        Msg::LangChanged(lang) => {
+            match lang.as_str() {
+                "en-US" => model.i18n.lang(Lang::en_US),
+                "de-DE" => model.i18n.lang(Lang::de_DE),
+                _ => model.i18n.lang(DEFAULT_LANG),
```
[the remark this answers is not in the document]
tokcum: The fact that all arms of a match statement have to return the same type pushed me to rethink this part of the code completely. I found the approach with FromStr appealing. What do you think?

examples/i18n/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+    args_female_pl.insert("userName", FluentValue::from("Anna"));
+    args_female_pl.insert("photoCount", FluentValue::from(5));
+    args_female_pl.insert("userGender", FluentValue::from("female"));
+    args_female_pl.insert("tabCount", FluentValue::from(7));
+    args_female_pl.insert("formal", FluentValue::from("false"));
+
+    div![
+        div![select![
+            attrs! {At::Name => "lang"},
+            langs,
```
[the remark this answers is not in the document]
tokcum: Hopefully some day I get used to this kind of thinking. I guess I need more practical experience. Done.

examples/i18n/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+    ]
+}
+
+fn translate(model: &Model, args: Option<&FluentArgs>, label: &str) -> String {
+    let msg = model
+        .i18n
+        .resource
+        .get_message(label)
+        .expect("Message doesn't exist.");
+    let mut errors = vec![];
```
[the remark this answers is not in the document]
tokcum: Me either. Maybe I have to read more on fluent. For now: Done.

examples/i18n/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+            )],
+            p![translate(model, None, "sync-dialog-title")],
+            p![translate(model, None, "sync-headline-title")],
+            p![translate(model, None, "sync-signedout-title")],
+        ],
+    ]
+}
+
+fn translate(model: &Model, args: Option<&FluentArgs>, label: &str) -> String {
+    let msg = model
```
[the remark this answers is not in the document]
tokcum: Done.

examples/i18n/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+        .i18n
+        .resource
+        .get_message(label)
+        .expect("Message doesn't exist.");
+    let mut errors = vec![];
+    let pattern = msg.value.expect("Message has no value.");
+
+    let value = model
```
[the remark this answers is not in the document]
tokcum: Refined the complete function. Done.

MartinKavik:
@tokcum I see you are working on it, click  `Re-request review` button when it's ready :)

tokcum:
@MartinKavik : ready. I like it more and more. :)

examples/i18n/src/i18n.rs
```diff
@@ -0,0 +1,96 @@
...
+    type Err = ();
+
+    fn from_str(s: &str) -> Result<Self, ()> {
+        match s {
+            "de-DE" => Ok(Self::de_DE),
+            "en-US" => Ok(Self::en_US),
+            _ => Err(()),
+        }
+    }
+}
```
MartinKavik, line 87: _"auto-derives std::str::FromStr on the enum."_ - [strum wiki](https://github.com/Peternator7/strum/wiki/Derive-EnumString). Perhaps it'll make the code cleaner and we already use `strum` (see also other `strum` macros, maybe there is more suitable one). You can try it, I don't have a strong opinion about `FromStr` vs `strum` macros.

examples/i18n/src/i18n.rs
```diff
@@ -0,0 +1,96 @@
...
+
+pub fn translate(i18n: &I18n, args: Option<&FluentArgs>, key: &str) -> String {
```
MartinKavik, line 89: I would move it to `I18` as its public method. 
You'll basically change only `translate(i18..` to  `translate(self: &Self` and it would be easily findable by "autocomplete exploration" - when user writes `model.i18n.` and see the list of all available methods. 
Also we get rid of `&` symbol - `translate(&model.i18n, ..)` vs `model.i18n.translate(..` and there would be less function arguments so it'll improve readability a little bit.
And my intuition tells me that this function doesn't belong here, below `Lang`, I would move it below `view` as the `view` helpers. However it makes more sense to throw it into `I18` because of reasons above.

examples/i18n/src/i18n.rs
```diff
@@ -0,0 +1,96 @@
...
+    let fluentmsg = i18n.resource.get_message(key).expect("get fluent message");
```
MartinKavik, line 90: `fluent_msg` please

examples/i18n/src/lib.rs
```diff
@@ -0,0 +1,123 @@
+use seed::{prelude::*, *};
+
+use fluent::fluent_args;
+use strum::IntoEnumIterator;
+
+mod i18n;
+use crate::i18n::{translate, I18n, Lang};
+use std::str::FromStr;
+
+mod resource;
```
MartinKavik, line 10: I suggest to regroup it a bit, something like:
```rust
use seed::{prelude::*, *};
use std::str::FromStr;

use fluent::fluent_args;
use strum::IntoEnumIterator;

mod i18n;
use i18n::{translate, I18n, Lang};   // <- you don't have to write "crate::" prefix
mod resource;
```

examples/i18n/src/i18n.rs
```diff
@@ -0,0 +1,96 @@
+use fluent::{FluentArgs, FluentBundle, FluentResource};
+use strum_macros::EnumIter;
+use unic_langid::LanguageIdentifier;
+
+use std::borrow::Borrow;
+use std::str::FromStr;
+
+use crate::resource::Resource;
```
MartinKavik, line 8: I think you can replace `crate` with `super` - your choice

examples/i18n/src/i18n.rs
```diff
@@ -0,0 +1,96 @@
+use fluent::{FluentArgs, FluentBundle, FluentResource};
+use strum_macros::EnumIter;
+use unic_langid::LanguageIdentifier;
+
+use std::borrow::Borrow;
+use std::str::FromStr;
```
MartinKavik, line 6: I would perhaps switch groups - i.e. imports from `std` would be first and external crates the second ; your choice

examples/i18n/src/resource.rs
```diff
@@ -0,0 +1,33 @@
+use std::collections::HashMap;
+
+#[derive(Debug)]
+pub struct Resource(HashMap<String, String>);
```
MartinKavik, line 4: My gut feeling tells me we don't need this struct at all. Try to eliminate it as the last step. If it's not possible I'll write more comments for it.

examples/i18n/src/i18n.rs
tokcum: Super. Done.

examples/i18n/src/i18n.rs
tokcum: Makes sense. Done.

examples/i18n/src/lib.rs
tokcum: To be consequent this means "use seed ..." is now in the second block. Done.

examples/i18n/src/i18n.rs
tokcum: Done. Also for ftlres -> ftl_res.

examples/i18n/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+enum Msg {
+    LangChanged(String),
+}
+
+fn update(msg: Msg, model: &mut Model, _: &mut impl Orders<Msg>) {
+    match msg {
+        Msg::LangChanged(lang) => {
+            model
+                .i18n
+                .set_lang(Lang::from_str(lang.as_str()).expect("supported language"));
```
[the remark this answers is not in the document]
tokcum: It's a learning opportunity. I feel like I understand the Trait concept a little bit better. Done.

examples/i18n/src/i18n.rs
tokcum: Agree, translate() is part of I18n. I should have moved it directly into I18n and not just into the module. I also rearranged arguments of translate(). First the key, than potential arguments. Done.

examples/i18n/src/i18n.rs
tokcum: Maybe I can get rid of Lang::id() when I use strum. I'll check ...

tokcum: Got rid of Lang::id() with strums AsRefStr. This cleans up several things.

examples/i18n/src/resource.rs
tokcum: We could name the FTL files like the Language Identifier and just load the FTL string when language is set or changed. This would eliminate module resource.rs completely. I'll try that ...

tokcum: If I understand correctly: include_str! expects a string literal so we can't build the path to the file dynamically.
I found [load_file](https://crates.io/crates/load_file) which allows that. For me it looks like that this supports requirement 5 "be able to provide translations by muggles.".

Ok, I head into this direction ...

tokcum: @MartinKavik Looks like load_file loads files at runtime but in the actual state of I18n I have to compile the translation into wasm. I thought about adding the FTL strings to every Lang variant or to I18n. But Lang and I18n just handle one language at a time while I would have to compile all FTL strings and keep track of which FTL string belongs to which language identifier.
Not sure how to handle that without complicating the example unnecessarily.

MartinKavik: I need to dive deeper and experiment a bit. I'll look at it tomorrow and probably create a PR to your branch. So please finish what you have.

Do we want lazy-load files with FTL strings?

tokcum: Thank you.

Nothing to commit. Working tree clean.

Regarding the delivery of the FTL strings: I think it depends. In case of a small web app supporting just a few languages it would be okay to compile everything into the wasm binary and load it all upfront. As soon as the web app grows or a lot of different languages have to be supported it would be nice to compile everything into wasm but load only the needed language. Is that possible in wasm?
One step further for a more dynamic multilingual approach where we can change the set of translations during runtime I think the web app will load the FTLs from the web server. I wouldn't implement this in this example. I think the focus should be to provide an example easy to adopt even for beginners. Having said that loading FTLs dynamically from a web server could be an additional example.

MartinKavik:
_Note:_ Blocked by https://github.com/tokcum/seed/pull/1

tokcum:
@MartinKavik I already allowed edit by maintainers so you should be able to change my pull request.

tokcum:
Ah, now I understand. You send a PR to my fork. Okay I’ll check that one.

MartinKavik:
> I already allowed edit by maintainers

1. There are some limitations - e.g. I think I can only add commits but cannot squash & force-push them.
1. I don't want to change the code under your hands.

tokcum:
@MartinKavik thank you very much for your support and time. I learned a lot from your comments and coaching.

MartinKavik:
@tokcum I'm glad you learned something. It was also good "user research" for me and Seed - I got many ideas for future Seed's docs.
Please the last step - could you rebase it and force-push (and maybe squash some commits)? There are some ugly commits,  I can't directly modify your branch and I can't rebase it to merge it into `master` now. Thank you!

MartinKavik:
Closing in favor of https://github.com/seed-rs/seed/pull/476

tokcum:
@MartinKavik thank you for merging. While git rebase was easy I stumbled with squashing commits.

MartinKavik:
> I stumbled with squashing commits

1. Shift + left mouse button
1. RMB
1. "Squash x commits"
1. Click message
1. Type a new commit message
1. Force push -> Yes
😄

PR #471, opened 2020-05-27
Replace `Blob::array_buffer` with `FileReader`

### LINKED ISSUE #470: WebsocketMessage.bytes() doesn't work on Safari
Hello,

`WebSocketMessage.bytes()` function doesn't work on safari and throws a runtime exception. This problem doesn't exist in Chrome or Firefox.

```
[Error] wasm-bindgen: imported JS function that was not marked as `catch` threw an error: – "getObject(arg0).arrayBuffer is not a function. (In 'getObject(arg0).arrayBuffer()', 'getObject(arg0).arrayBuffer' is undefined)↵↵Stack…"
	(anonymous function) (package.js:224)
	wasm-stub
	<?>.wasm-function[web_sys::features::gen_Blob::Blob::array_buffer::h6035f6f3793b884c]
	<?>.wasm-function[seed::browser::web_socket::message::WebSocketMessage::bytes::{{closure}}::h4555098a4d829d6b]
	<?>.wasm-function[<std::future::GenFuture<T> as core::future::future::Future>::poll::h15af837ff880ad89]
	<?>.wasm-function[std::future::poll_with_tls_context::hecb607a7f8c88198]
	<?>.wasm-function[client::chat::update::{{closure}}::h747c3fe65ead2501]
	<?>.wasm-function[<std::future::GenFuture<T> as core::future::future::Future>::poll::hc29e47b33fd89f21]
	<?>.wasm-function[<futures_util::future::future::map::Map<Fut,F> as core::future::future::Future>::poll::h2a1c1f0aec3006e2]
	<?>.wasm-function[<futures_util::future::future::Map<Fut,F> as core::future::future::Future>::poll::h6dafe6092c28d87f]
	<?>.wasm-function[wasm_bindgen_futures::task::singlethread::Task::run::hd2ae8a7fce6033b4]
	<?>.wasm-function[wasm_bindgen_futures::queue::QueueState::run_all::h4f7092ed55373ca7]
	<?>.wasm-function[wasm_bindgen_futures::queue::Queue::new::{{closure}}::hb67de4d16789d1b1]
	<?>.wasm-function[<dyn core::ops::function::FnMut<(A,)>+Output = R as wasm_bindgen::closure::WasmClosure>::describe::invoke::h727a037c3544a9ea]
	wasm-stub
	_dyn_core__ops__function__FnMut__A____Output___R_as_wasm_bindgen__closure__WasmClosure___describe__invoke__h727a037c3544a9ea
	__wbg_adapter_31 (package.js:273:135)
	real (package.js:249)
	promiseReactionJob
```

Function `bytes()`(message.rs)  at line 52 uses `blob.array_buffer()` which is currently not supported by safari.
The quick fix for this issue would be to use `FileReader.readAsArrayBuffer`.

tkubicz:
This PR replaces the `Blob:array_buffer` usage with the FileReader. This change fixes #470.

This is my first PR to seed, and the first time that I had anything to do with wasm_bindgen so I would really appreciate a review.

### DIFF
--- a/examples/websocket/src/client.rs
+++ b/examples/websocket/src/client.rs
@@ -12,6 +12,7 @@ struct Model {
     sent_messages_count: usize,
     messages: Vec<String>,
     input_text: String,
+    input_binary: String,
     web_socket: WebSocket,
     web_socket_reconnector: Option<StreamHandle>,
 }
@@ -25,6 +26,7 @@ fn init(_: Url, orders: &mut impl Orders<Msg>) -> Model {
         sent_messages_count: 0,
         messages: Vec::new(),
         input_text: String::new(),
+        input_binary: String::new(),
         web_socket: create_websocket(orders),
         web_socket_reconnector: None,
     }
@@ -47,12 +49,15 @@ fn create_websocket(orders: &impl Orders<Msg>) -> WebSocket {
 enum Msg {
     WebSocketOpened,
     MessageReceived(WebSocketMessage),
+    BytesReceived(Vec<u8>),
     CloseWebSocket,
     WebSocketClosed(CloseEvent),
     WebSocketFailed,
     ReconnectWebSocket(usize),
     InputTextChanged(String),
+    InputBinaryChanged(String),
     SendMessage(shared::ClientMessage),
+    SendBinaryMessage(shared::ClientMessage),
 }
 
 fn update(msg: Msg, mut model: &mut Model, orders: &mut impl Orders<Msg>) {
@@ -63,9 +68,22 @@ fn update(msg: Msg, mut model: &mut Model, orders: &mut impl Orders<Msg>) {
         }
         Msg::MessageReceived(message) => {
             log!("Client received a message");
-            model
-                .messages
-                .push(message.json::<shared::ServerMessage>().unwrap().text);
+
+            if message.contains_text() {
+                model
+                    .messages
+                    .push(message.json::<shared::ServerMessage>().unwrap().text);
+            } else {
+                orders.perform_cmd(async move {
+                    let bytes = message.bytes().await;
+                    bytes.map(Msg::BytesReceived).ok()
+                });
+            }
+        }
+        Msg::BytesReceived(bytes) => {
+            log!("Client received binary message");
+            let msg: shared::ServerMessage = rmp_serde::from_slice(&bytes).unwrap();
+            model.messages.push(msg.text);
         }
         Msg::CloseWebSocket => {
             model.web_socket_reconnector = None;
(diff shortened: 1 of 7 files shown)


MartinKavik:
Thanks for the PR! 

I was looking at it and it looks good, I would add only a few comments.
However I found [this code](https://github.com/rustwasm/gloo/blob/2c9e776701ecb90c53e62dec1abd19c2b70e47c7/crates/file/src/file_reader.rs#L209) that directed me to function [read_as_bytes](https://github.com/rustwasm/gloo/blob/2c9e776701ecb90c53e62dec1abd19c2b70e47c7/crates/file/src/file_reader.rs#L209). Seed's WebSocket wrapper already uses `gloo_file::Blob` so I can naively imagine that we can just write `read_as_byte(&blog).await` and done - but I don't know if there are some methods that don't work on Safari, you have better knowledge about it than me now => in that case we would return back to your implementation.

Also we should add unit test or perhaps ideally extend WebSocket example in the Seed repo to use also method`bytes` - would you like to do it? I think you are more experienced than me with byte WebSocket messages so I would be glad.

MartinKavik:
_Note:_ The latest `wasm-bindgen` supports `Blob.stream` (see [README](https://github.com/rustwasm/wasm-bindgen/blob/master/CHANGELOG.md#added)), but it isn't supported by all browsers, too.

tkubicz:
HI Martin,

You were right, using read_as_bytes from gloo is much better approach and works just fine on every browser tested by me (Chrome, Firefox, Safari and Opera). I have also extended the `web socket` example to show the usage of binary messaging.

I'm having trouble with implementing tests for this function, because I have no idea how to convert js_sys::Uint8Array to JsValue and create a blob directly in rust.

MartinKavik:
Nice, thanks! I'll check how it works on Windows and try to do some quick review.

> I have no idea how to convert js_sys::Uint8Array to JsValue and create a blob directly in rust.

Does [`AsRef<JsValue> for Uint8Array`](https://rustwasm.github.io/wasm-bindgen/api/js_sys/struct.Uint8Array.html#impl-AsRef%3CJsValue%3E) or [`From<Uint8Array> for JsValue`](https://rustwasm.github.io/wasm-bindgen/api/js_sys/struct.Uint8Array.html#impl-From%3CUint8Array%3E) help?
`gloo_file` has [Blob::new](https://docs.rs/gloo-file/0.1.0/gloo_file/struct.Blob.html#method.new) and [Blob::new_with_option](https://docs.rs/gloo-file/0.1.0/gloo_file/struct.Blob.html#method.new_with_options). I assume that gloo's File and native blobs and files has similar methods. Also there are `From` implementation for casting among all types.

src/browser/web_socket.rs
```diff
@@ -50,6 +51,7 @@ pub enum WebSocketError {
     SendError(JsValue),
     SerdeError(serde_json::Error),
     PromiseError(JsValue),
+    FileRedaerError(FileReadError),
```
MartinKavik, line 54: Typo

examples/websocket/src/client.rs
```diff
@@ -63,9 +68,22 @@ fn update(msg: Msg, mut model: &mut Model, orders: &mut impl Orders<Msg>) {
...
+
+            if message.contains_text() {
+                model
+                    .messages
+                    .push(message.json::<shared::ServerMessage>().unwrap().text);
+            } else {
+                orders.perform_cmd(async move {
+                    let bytes = message.bytes().await;
+                    bytes.map(Msg::BytesReceived).ok()
+                });
```
MartinKavik, line 80: OT a bit:
Do you think that `update` function should be `async` or that we should add blocking function like `orders.perform_cmd_sync` for similar cases? Or is the current Elm-like API enough / better?

tkubicz: Async `update` function would be really great, but I assume that it would require quite a lot of development and it's not really a necessity. Using messaging API (orders) is enough - introducing blocking function is not really necessary here in my opinion.

tkubicz:
Ok, I've corrected the typo, added changelog entry and unit test to check if obtaining bytes really work.

Issue #433, opened 2020-04-27
Suggestions for Url-Api


power-fungus:
I did a review of the url-api. My motivation is that I did not understand how this module is intended to be used and saw an oppertunity to improving discoverability and perceived consistency of the api.

I believe I understand the intention behind the provided functionality. I understand its usecases for routing and I do not want to remove any features from the module.

## Name the new parts of the url

The Url struct is not the *typical* url. It represents more than just the url (it has extra features for routing). All parts of a typical url are already named corretly (path, hash, search). The "Base path" is a concept introduced by the routing-functionaliyt and is implicitly named. 

I suggest embracing concept of the base-path. In current terms, it is the parts of the path that are "behind" of the url. The functions `to_base_url` and `skip_base_path` hint at that fact.
If there is a "base-path", there should also be a name for the rest. I see two possibilities:

* `path = base-path + head-path`. The name `head_path` is the best name I could came up with, it is to be bikeshed.
* `absolute-path = base-path + path`. This would be a breaking change and is quite disruptive.

The following describe suggested changes to functions:

### to_base_url

This method clones the url and truncates the path. Suggestions:

* The method should take `&mut self` to avoid cloning inside the library - users can still clone if they like to.
* The method does not truncate the hash, which it should. Conceptually, the hash comes "after" the path.

### next_path_part

Instead of "Advances the internal path iterator", the documentation of this method desccribe somthing along the lines of: "the next part of the url is moved to the base-url section of the path".

### return string slices

* implement `base_path(&self) -> &[String]`.
* `remaining_path_parts` should be `head_path(&self) -> &[String]` (no `mut`, return slice instead of vec)

### consistent subtitle

**URL used for routing** can serve as a subtitle for seed's concept of url and could to be communicated more. This highlights that this is not the *typical* url-struct (just data) and serves the specific purpose. It should also documented that the url-path is split into 

* module-level description should start with that
* struct-description should start with that (it already does)
* guide-documentation

### const DUMMY_BASE_URL

This is a const to "example.com". This seems a bit odd, currently the url-struct represents relative urls (which is nice). Having this const in the framework does not make much sense.

## struct_urls macro seems off

It generates a newtype around a base-url. I understand that the custom methods (like `home()`) can be useful, but I doubt that this is the best abstraction.
I suggest removing the `struct_urls!` macro and having the users require to store the base-path inside the model. The functions like `home()` can still be implemented somewere.

## reorder URL-methods

[...]

Issue #383, opened 2020-03-09
Url/Router - new API design


MartinKavik:
**Motivation**

- Reduce boilerplate.
- [Slice patterns](https://thomashartmann.dev/blog/feature(slice_patterns)/) will be stabilized in the next Rust version (`1.42.0`).
- Seed should take into account base path (https://github.com/seed-rs/seed/issues/369).
- User should be able to get important data from the `Url` - e.g. domain or protocol.
- User should be able to pass the `Url` to services/APIs like `fetch`.

**Inspiration**

- https://github.com/reasonml/reason-react/blob/master/docs/router.md
- https://package.elm-lang.org/packages/elm/browser/latest/Browser-Navigation
- https://package.elm-lang.org/packages/elm/url/latest/Url-Parser
- https://package.elm-lang.org/packages/sporto/hop/latest/Hop
- http://codeloveandboards.com/blog/2019/06/06/dynamic-base-path-for-an-elm-spa/
- https://github.com/elm/url/issues/14
- https://github.com/mint-lang/mint-realworld/blob/master/source/Routes.mint
- https://reacttraining.com/react-router/web/guides/quick-start
- https://router.vuejs.org/
- https://github.com/ItalyPaleAle/svelte-spa-router
- https://github.com/jorgegorka/svelte-router
- https://scrimba.com/playlist/pMvYcg

TatriX:
What do you think about giving an iterator instead of custom methods?
Instead of
```rust
match url.next_path_part() {
   None => Page::Home,
   Some("report") => Page::Report(page::report::init(url)),
   _ => Page::Unknown(url),
}
```
something like
```rust
match url.into_path_iter().next() {
   None => Page::Home,
   Some("report") => Page::Report(page::report::init(url)),
   _ => Page::Unknown(url),
}
```

I used to `...collect::<Vec<&str>>` an `Url` so I can
```rust
match segments {
            [""] => Screen::Home,
            ["about"] => Screen::About,
            ["profile"] => Screen::Profile,
            ["signin"] => Screen::SignIn,
            ["signup"] => Screen::SignUp,
...
```

flosse:
> What do you think about giving an iterator instead of custom methods?

:+1: 

what about
```rust
url.path().iter().next()
```
or
```rust
url.path().into_iter().next()
```

Instead of

> ```rust
> url.into_path_iter().next()
> ```

TatriX:
[...]

PR #472, opened 2020-05-30
Url API changes

AlterionX:
Mostly cosmetic changes, reorganization, and edits to the documentation.

The biggest change here is probably with the `UrlSearch` struct's `FromIterator` impl as well as how `UrlSearch` has been extracted into `seed::browser::url::search` (but reexported).

The documentation part was left out, mostly because I'm not entirely sure what would qualify as "most common" use cases.

Should handle most of #433, but all changes are up for discussion.

Also related to #383.

### DIFF
--- a/examples/url/src/lib.rs
+++ b/examples/url/src/lib.rs
@@ -77,7 +77,7 @@ fn view(model: &Model) -> Node<Msg> {
                         .set_path(&["ui", "a", "b", "c"])
                         .set_search(vec![
                             ("x", vec!["1"])
-                        ].iter().collect())
+                        ].into_iter().collect())
                         .set_hash("hash")
                         .go_and_load()
                 })
--- a/examples/url/src/lib.rs
+++ b/examples/url/src/lib.rs
@@ -75,9 +75,9 @@ fn view(model: &Model) -> Node<Msg> {
                 ev(Ev::Click, |_| {
                     Url::new()
                         .set_path(&["ui", "a", "b", "c"])
-                        .set_search(UrlSearch::new(vec![
+                        .set_search(vec![
                             ("x", vec!["1"])
-                        ]))
+                        ].iter().collect())
                         .set_hash("hash")
                         .go_and_load()
                 })
--- a/examples/url/src/lib.rs
+++ b/examples/url/src/lib.rs
@@ -32,10 +32,10 @@ impl Model {
         Self {
             base_path,
             initial_url: url.clone(),
-            base_url: url.to_base_url(),
-            next_path_part: url.next_path_part().map(ToOwned::to_owned),
+            base_url: url.clone().truncate_relative_path(),
+            next_path_part: url.pop_relative_path_part().map(ToOwned::to_owned),
             remaining_path_parts: url
-                .remaining_path_parts()
+                .consume_relative_path()
                 .into_iter()
                 .map(ToOwned::to_owned)
                 .collect(),
--- a/src/browser/url/search.rs
+++ b/src/browser/url/search.rs
@@ -1,6 +1,6 @@
+use super::Url;
 use serde::{Deserialize, Serialize};
 use std::{borrow::Cow, collections::BTreeMap, fmt};
-use super::Url;
 
 #[allow(clippy::module_name_repetitions)]
 #[derive(Default, Debug, Clone, Serialize, Deserialize, PartialEq)]
@@ -138,13 +138,14 @@ impl From<web_sys::UrlSearchParams> for UrlSearch {
 }
 
 impl<K, V, VS> std::iter::FromIterator<(K, VS)> for UrlSearch
-    where
-        K: Into<String>,
-        V: Into<String>,
-        VS: IntoIterator<Item = V>,
+where
+    K: Into<String>,
+    V: Into<String>,
+    VS: IntoIterator<Item = V>,
 {
     fn from_iter<I: IntoIterator<Item = (K, VS)>>(iter: I) -> Self {
-        let search = iter.into_iter()
+        let search = iter
+            .into_iter()
             .map(|(k, vs)| {
                 let k = k.into();
                 let v: Vec<_> = vs.into_iter().map(Into::into).collect();
@@ -159,17 +160,16 @@ impl<K, V, VS> std::iter::FromIterator<(K, VS)> for UrlSearch
 }
 
 impl<'a, K, V, VS> std::iter::FromIterator<&'a (K, VS)> for UrlSearch
-    where
-        K: 'a,
-        &'a K: Into<String>,
-        V: Into<String>,
-        VS: 'a,
-        &'a VS: IntoIterator<Item = V>,
+where
+    K: 'a,
+    &'a K: Into<String>,
+    V: Into<String>,
+    VS: 'a,
+    &'a VS: IntoIterator<Item = V>,
 {
     fn from_iter<I: IntoIterator<Item = &'a (K, VS)>>(iter: I) -> Self {
-        iter.into_iter().map(|(k, vs)| (
-            k.into(),
-            vs.into_iter(),
-        )).collect()
+        iter.into_iter()
+            .map(|(k, vs)| (k.into(), vs.into_iter()))
+            .collect()
     }
 }
--- a/src/browser/url/search.rs
+++ b/src/browser/url/search.rs
@@ -1,0 +1,159 @@
+use serde::{Deserialize, Serialize};
+use std::{borrow::Cow, collections::BTreeMap, fmt};
+
+use super::Url;
+
+// ------ UrlSearch ------
+
+#[allow(clippy::module_name_repetitions)]
+#[derive(Default, Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct UrlSearch {
+    search: BTreeMap<String, Vec<String>>,
+    pub(super) invalid_components: Vec<String>,
+}
+
+impl UrlSearch {
+    /// Makes a new `UrlSearch` with the provided parameters.
+    ///
+    /// # Examples
+    ///
+    /// ```rust,no_run
+    /// UrlSearch::new(vec![
+    ///     ("sort", vec!["date", "name"]),
+    ///     ("category", vec!["top"])
+    /// ])
+    /// ```
+    pub fn new<K, V, VS>(params: impl IntoIterator<Item = (K, VS)>) -> Self
+    where
+        K: Into<String>,
+        V: Into<String>,
+        VS: IntoIterator<Item = V>,
+    {
+        let mut search = BTreeMap::new();
+        for (key, values) in params {
+            search.insert(key.into(), values.into_iter().map(Into::into).collect());
+        }
+        Self {
+            search,
+            invalid_components: Vec::new(),
+        }
+    }
+
+    /// Returns `true` if the `UrlSearch` contains a value for the specified key.
+    pub fn contains_key(&self, key: impl AsRef<str>) -> bool {
+        self.search.contains_key(key.as_ref())
+    }
+
+    /// Returns a reference to values corresponding to the key.
+    pub fn get(&self, key: impl AsRef<str>) -> Option<&Vec<String>> {
+        self.search.get(key.as_ref())
+    }
+
+    /// Returns a mutable reference to values corresponding to the key.
+    pub fn get_mut(&mut self, key: impl AsRef<str>) -> Option<&mut Vec<String>> {
+        self.search.get_mut(key.as_ref())
+    }
+
+    /// Push the value into the vector of values corresponding to the key.
+    /// - If the key and values are not present, they will be crated.
+    pub fn push_value<'a>(&mut self, key: impl Into<Cow<'a, str>>, value: String) {
+        let key = key.into();
+        if self.search.contains_key(key.as_ref()) {
+            self.search.get_mut(key.as_ref()).unwrap().push(value);
+        } else {
+            self.search.insert(key.into_owned(), vec![value]);
+        }
+    }
+
+    /// Inserts a key-values pair into the `UrlSearch`.
+    /// - If the `UrlSearch` did not have this key present, `None` is returned.
+    /// - If the `UrlSearch` did have this key present, old values are overwritten by new ones,
+    /// and old values are returned. The key is not updated.
+    pub fn insert(&mut self, key: String, values: Vec<String>) -> Option<Vec<String>> {
+        self.search.insert(key, values)
+    }
+
+    /// Removes a key from the `UrlSearch`, returning values at the key
+    /// if the key was previously in the `UrlSearch`.
+    pub fn remove(&mut self, key: impl AsRef<str>) -> Option<Vec<String>> {
+        self.search.remove(key.as_ref())
+    }
+
+    /// Gets an iterator over the entries of the `UrlSearch`, sorted by key.
+    pub fn iter(&self) -> impl Iterator<Item = (&String, &Vec<String>)> {
+        self.search.iter()
+    }
+
+    /// Get invalid components.
+    ///
+    /// Undecodable / unparsable components are invalid.
+    pub fn invalid_components(&self) -> &[String] {
+        &self.invalid_components
+    }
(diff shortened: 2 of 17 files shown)


AlterionX:
I need to find a linux machine somewhere to run `cargo make test` at some point, or figure out how to make WSL 2 run the tests -- as it is, the tests are failing due to lack of packages, incompatability issues, etc.

MartinKavik:
> I need to find a linux machine somewhere to run cargo make test

Why? What's your problem and OS?

AlterionX:
Windows, WSL 2
I'm probably just missing a few packages, but it's late and I wanted to put it off for tomorrow.

examples/url/src/lib.rs
```diff
@@ -75,9 +75,9 @@ fn view(model: &Model) -> Node<Msg> {
                 ev(Ev::Click, |_| {
                     Url::new()
                         .set_path(&["ui", "a", "b", "c"])
-                        .set_search(UrlSearch::new(vec![
+                        .set_search(vec![
                             ("x", vec!["1"])
-                        ]))
+                        ].into_iter().collect())
```
MartinKavik, line 80: I don't know if it's better or worse: two calls + hidden type VS explicit type `UrlSearch` that needs to be imported.
Do you think we can write only `.set_search(vec![("x", vec!["1"])])` or `set_search(UrlSearch::new..)` and don't complicate `set_search` parameter too much?

AlterionX:
Apparently some people agree about clippy's filter_map lint. https://github.com/rust-lang/rust-clippy/issues/3188

Would it be okay if I disabled it?

src/browser/url/search.rs
```diff
@@ -0,0 +1,175 @@
...
+where
+    K: 'a,
+    &'a K: Into<String>,
+    V: Into<String>,
+    VS: 'a,
+    &'a VS: IntoIterator<Item = V>,
+{
+    fn from_iter<I: IntoIterator<Item = &'a (K, VS)>>(iter: I) -> Self {
+        iter.into_iter()
+            .map(|(k, vs)| (k.into(), vs.into_iter()))
```
MartinKavik, line 172: Could you use more expressive names? I.e. `k` -> `key` and `vs` => `values`?

MartinKavik:
> Windows, WSL 2

I use Windows without WSL, with git-bash instead of cmd/ps and it works without problems.

> Apparently some people agree about clippy's filter_map lint.

I've already disabled it somewhere, I won't cry if you do that.

MartinKavik:
- `url.to_base_url()` -> `url.clone().truncate_relative_path()`
   - `relative_path` implies that there is also an `absolute_path` but `Url` uses only relative paths from the user point of view. Also I'm afraid it will be even more confusing once we add support for real relative and absolute paths.
  - Alternative names: `url.clone().trim_path()`, `.trim_path_end()`, `.shrink_path_to_base()`, ..
  - Or leave it as is - is domain-specific naming and implicit cloning bad? And prefix `to_` also implies that `to_base_url` returns `Url`; I would expect that `truncate_relative_path` returns `()`. 

- `url.next_path_part()` -> `url.pop_relative_path_part()`
   - `pop` usually removes the last element from collections - we don't remove anything here and it should be named `pop_front` to respect conventions => I would leave `.next_path_part` as is.

- `self.base_url().add_path_part(LOGIN)` -> `self.base_url().push_path_part(LOGIN)`
   - I agree, `push` makes more sense than `add`.

- `url.remaining_path_parts().as_slice()` -> `url.consume_relative_path().as_slice()`
  - What do you think about `collect` instead of `consume`? This method only borrows, doesn't consume path parts. So it would change to `.collect_path` or `.collect_remaining_path` to be super expressive.

(cc @power-fungus)

MartinKavik:
Just idea for future explorations - [Cursors](https://doc.rust-lang.org/std/collections/linked_list/struct.CursorMut.html) - perhaps it would help with implementation.

AlterionX:
Thanks for the comments. Next, I've been thinking about what you said, and wanted to write something comprehensive and well thought out. I'm sorry for the wall of text.

I think we need to talk a bit about what the two parts of the path should be named -- ideally something that speaks more to the intended purpose of `Url` than `remaining_path`. Also, it's not immediately obvious what `remaining` means--the word `remaining` doesn't seem to pertain to the `Url`. I agree that `relative_path` is confusing, but I think it's more declarative than `remaining_path` and better than `head_path` as suggested, or `tail_path` since there is no `head_path`. I'm not  I don't want to rename `base_path` to `head_path` either, since I think `base_path` speaks extremely clearly to what it's meant for. The names `parent_path` and `child_path` also popped up, but I don't like those either. As such, I've been unable to come up with a good name for that part of the path. Taking this into account, I have added a small snippet to the docs of functions pertaining to the base and remaining paths to clarify their meaning.

I also think that `Url` should be named something different like `UrlPath`/`WebPath`/`SitePath` (not `AppPath` since it's not directly related to the application), since it is both more than and less than `web_sys::Url`. Also, it'll avoid confusing people if someone sees `Url` and assumes it's an actual `Url`. The reason I avoid untyped languages is due to their uncertainty about what data ends up where. The name `Url` in this case makes me uncertain about it's difference with `web_sys::Url`. I think that doing this in the code would make it more obvious to everyone involved about how they're different.

Sidenote: The second part's suggested change may mitigate the confusion from the first part, if we go down that route. However, I still think that there should be a better name than what we have come up with so far.

Third, regarding the `consume_*` methods or `Url`, I was also trying to come up with a better name. The key part here that I was attempting to highlight is that these two methods "eat up" the rest of the relative path, shunting their path parts into the base path. I think that `collect_remaining_path` doesn't exactly imply that this happens. I expect `collect_*` to create a `Vec` that's populated with whatever `*` is, but not change the `Url`. (My brain thinks a little differently for `Iterator`s, hence the apparent inconsistency there. Another reason I wanted to avoid collect, because it's almost but not quite an iterator.) To that effect I've written a new method -- `relative_path` -- that would do what I expect collect does, short of allocating a `Vec`.

Fourth, regarding `to_base_path`/`truncate` -- I wanted to make it return nothing and just truncate the path, but was worried it might be a big change. That said, I will change it to return `()` instead soon, if that's okay. I had issue with the `to_base_path` method since it didn't consume the original `Url`. Another part was that I wanted people to avoid thinking that `truncate` acted on the original `Url`, so I tried to make it obvious that `truncate` needed ownership to act, but that's honestly not that important.

Fifth, regarding the change to `set_search`, I think that making the conversion here noisy is better, since I would personally prefer a macro for dictionary construction than implicitly converting a `Vec` into a `BTreeMap`, doubly so since it's not immediately clear that a `Vec` should map to `BTreeMap`. That said, we can use the `UrlSearch::from_iter` method, but that would require importing the `FromIterator` trait (see [`FromIterator`'s documentation](https://doc.rust-lang.org/std/iter/trait.FromIterator.html) for more details regarding `from_iter` and how it's not typically called by itself). It wouldn't be hard to write another method to avoid importing `FromIterator`, but I still think that such a function should be avoided. In conclusion, I think the real solution is to write a macro for this particular case as from the [`maplit` crate](https://crates.io/crates/maplit), like so:

```rs
url_search! {
    "query_param0" => ["data", "thing"],
    "query_param1" => [],
    "query_param2" => "other thing",
}
```

Lastly, regarding `pop_*`, I was thinking of a longer method name, but was worried it would take too long. Thinking on it a bit more, I suppose I don't like `pop_*` after all, but I also don't like `next_path_part` since I don't think that `Url` should act as if it has an internal iterator--again, we would be mixing concepts that don't usually go together. Now if it just didn't fit together, I wouldn't have much of an objection, but this paired with the fact that the struct is more than just an iterator makes me feel like it's rather confusing what, exactly, is going on. Having said that, I was thinking of `shift_next_path_part_to_base` since `next_path_part` sounds more like a getter outside the context of an iterator, but as mentioned, it felt like it was too long. Perhaps we can break this up into two methods or rename it so that the function's purpose as well as what it does is a little clearer.

Another sidenote: I left a bunch of `TODO`s, could you take a look and see which ones seem reasonable? They're there since this was supposed to be closer to a discussion.

MartinKavik:
I've just read it quickly so keep it in mind while reading notes below:

> Also, it'll avoid confusing people if someone sees Url and assumes it's an actual Url.
  - There was/is plan to make it full-fledged Url - it should replace `web_sys::Url`. Just like we have wrappers for fetch and websockets.

>  I expect collect_* to create a Vec that's populated with whatever * is, but not change the Url.
  - Well, I think of Url as a special Iterator - basically a container for multiple iterators. So I expect it changes Url a bit - just like standard Iterator.

> That said, I will change it to return () instead soon, if that's okay.
  - I think it would decrease developer experience a bit because you can't call it and store the new Url into your Model on one line (?).

>  I had issue with the to_base_path method since it didn't consume the original Url.
  - Yes, that's why the prefix is `to_` and not `into_` to respect naming conventions and make it explicit.

> Fifth, regarding the change to set_search
  - I would design it simple as possible - `UrlSearch::new` + `url_search!` (as you suggested) - basically like `Vec::new` + `vec![]` or other collections and that `maplit` crate.

> next_path_part since I don't think that Url should act as if it has an internal iterator--again, we would be mixing concepts that don't usually go together
  - Well, how I said, it's basically Iterator for me and it should be easy to grasp the concept because Rust developers already know Iterators. I think we won't agree on this - perhaps we'll need more opinions.

AlterionX:
Okay, so let's backtrack a little here, since I think what I'm trying to convey is getting a little lost in the details.

Can we agree that a struct should have one "role", so to speak? The single responsibility principle from OOP, as it were. `Url` as it is now has two separate roles: it attempts to be an iterator as well as represent the url. From it's name, I only notice the second, not the first. It requires some deep delving in the docs to conclude the first purpose of `Url`. It says "Url used for routing," but that doesn't explain the internal state or what it's supposed to represent.

This leads to what I believe to be the most important part of my response earlier: I want to rename `Url` as well as name the relevant components. Here's the paragraph:

> I also think that Url should be named something different like UrlPath/WebPath/SitePath (not AppPath since it's not directly related to the application), since it is both more than and less than web_sys::Url. Also, it'll avoid confusing people if someone sees Url and assumes it's an actual Url. The reason I avoid untyped languages is due to their uncertainty about what data ends up where. The name Url in this case makes me uncertain about it's difference with web_sys::Url. I think that doing this in the code would make it more obvious to everyone involved about how they're different.

Before quibbling about the methods and whatnot (not to say that these details are trivial, just secondary to the primary point), I wanted to establish a clearer picture of what the purpose of the `Url` struct is and is not.

I think we can arrive at an agreement if we can decide on a new name for `Url` that better captures what it is or split `Url`'s functionality into several components. From your description, however, I'm a little worried about `Url` adopting even more varied methods and it's purpose (read responsibility) getting more muddled than it is now.

MartinKavik:
> Can we agree that a struct should have one "role", so to speak? The single responsibility principle from OOP, as it were. Url as it is now has two separate roles: it attempts to be an iterator as well as represent the url. 

I'm aware that `Url`:
  1. Contains "custom" Iterators.
  1. Doesn't represent `Url` correctly (it doesn't support absolute path now).
  1. Is mixed with routing (e.g. you can push route into the browser history through `Url` methods).

And that everything above is bad on a paper.

So yes I also
> wanted to establish a clearer picture of what the purpose of the Url struct is and is not.

But all my attempts to write idiomatic Rust failed when I tried to integrate them into bigger examples and real-world apps.

>  It says "Url used for routing," but that doesn't explain the internal state or what it's supposed to represent short of the fact that routing is somehow involved and it's a url.

I think I just copy/pasted old comment because I didn't have time to write something better and I didn't know all problems and patterns to write it properly.

> I think we can arrive at an agreement if we can decide on a new name for Url that better captures what it is or split Url's functionality into several components. From your description, however, I'm a little worried about Url adopting even more varied methods and it's purpose (read responsibility) getting more muddled than it is now.

- The goal was to use `Url` as a wrapper for `web_sys::Url`. If the goal is still valid, then renaming doesn't make sense.
- DX and usage in examples and especially in bigger real-world apps have priority. I don't care about theory and a bit steeper learning curve if it helps people in the long run. And there were some positive feedbacks on Seed chat. So - I don't want to change `Url` until it's clear that it helps with writing apps and we have proves in examples.

So.. I agree with you that there are some Rust/OOP/SoC issues, but there are considered trade-offs. So please show us how your changes improve user life while he is writing more complex apps if you want to change API.

power-fungus:
I am happy to see that @AlterionX 's  suggestions and changes go in a direction I quite like. 

Some thoughts:

* Having relative-urls (also relative to the app-base-path) is a surprisingly useful feature for routung. The plan to also support absolute urls (for CORS-requests) should not get in the way of the routing api (by making absolute paths optional).
* Regarding iterators:
  * conceptually, `web_sys::Url` is a stateless struct, adding state is quite a surprise
  * using iterator-getters (`url.path().into_iter()`) instead of mutable state *should work* in all cases I can think of and is certainly the more expected api for newcomers.
  * If the iterator-nature of `Url` should be kept, the name should represent 
* is the splitting of the hash into parts really beneficial? Is the character `/` the obvious segment seperator? Does any other framework set `/` as hash-separator? If not, I suggest returning to `hash() -> &str`
* @AlterionX in [your comment](https://github.com/seed-rs/seed/pull/472#issuecomment-637742484), you discuss naming preferences. Did you get to any conclusions? Personally I would prefer one of 
  * `<base_path>/<head_path>` (`tail_path` would also work)
  * `<base_path>/<path>` (this one might be a bit unexpected, since `path()` automatically skips the base.

MartinKavik:
> using iterator-getters (url.path().into_iter()) instead of mutable state should work in all cases

It doesn't work when you want to handle each path part and other url components in the associated page module. See some examples - e.g. [pages](https://github.com/seed-rs/seed/tree/master/examples/pages). Better alternative welcome.
And motivation for handling in page modules - one central router becomes messy when you have many pages. 

> is the splitting of the hash into parts really beneficial? Is the character / the obvious segment seperator? Does any other framework set / as hash-separator? If not, I suggest returning to hash() -> &str

There are two methods - `hash()` returns `Option<&String>` (it basically corresponds with your `hash()->&str`) and the second one - `hash_path()` - returns `&[String]` (the same output like `path()`) and it's for users who need hash routing because of conflicts with backend endpoints.

power-fungus:
looking at the pages example, one could change the `init` functions to be one of:

* `init(url_path: mut impl Iterator<Item=&String>)`, instead of `url.next_path_part()` one could call `url_path.next()`
* `init(url_path: &[&String])`, instead of `url.next_path_part()` one could call `url_path.first()` and instead of passing the same url struct to the child init-function, one could pass `&url_path[1..]` instead.

If any other additional information about the rest of the url would be required, one would need to add additional parameters - with the additional benefit for documentation that these parameters need to be named.

MartinKavik:
@power-fungus Yeah, it's one reasonable alternative. 
I didn't want to go this way because:
- I can't predict what users will  need - it would be sad if it would force them to pass multiple iterators / arrays / strings. Also it would be quite error-prone because you basically pass only Strings/&strs (lost expressive/domain-specific names).
- I don't know if it would improve link building or make it worse.
- How would it work with base paths?
- Types like `mut impl Iterator<Item=&String>` would reduce readability.

However I can be wrong and this API would be better in practice.

power-fungus:
I agree about not predicting the usecases of users. It has proven to be a useful heuristic to favor designs that do less but are more composable. seed's view system is a good example of this.

I made a mistake: `mut` is part of the pattern, not the type. It should go `init(mut url_path: impl Iterator<Item=...>)`. Thinking about mutability is a first-class concern in all Rust programming and is common enough to not hinder readability.
Personally, I would prefer the slice-passing-option, but there are also other options (in the spirit of composability)

Link building is a different consideration (in my view). Probably, it is fine to use the current `push_path_part` on the base-url.

A hybrid approach exists: `pop_relative_path_part` (as in this PR) that changes the part of the path that is considered "base". I would not consider this advance an iterator or cursor, but rather as a mutation of the url-struct (similar to a write to a collection).

flosse:
This PR seems to be outdated and there was no chance since 2020 so I'll close it now. Feel free to reopen it!

PR #491, opened 2020-06-30
examples/jwt

eopb:
work in progress


ethanboxx:
I have added some notes on security to the readme.

The original reason I used redirects was to provide good compatibility with form actions. I attempted to use fetch but I was unable to get the client to "see" the cookies. See https://github.com/ethanboxx/seed/pull/2 . Maybe I am missing something?

The redirects are very noticeable in this example because the page does not change much but, in a normal case, I don't think it would be noticeable. As far as I can tell other sites I have inspected behave that way. I think users expect page reload behaviour on login.

MartinKavik:
> I attempted to use fetch but I was unable to get the client to "see" the cookies.

The next time I would rather learn how to fly Boeing than set up cookies... see PR https://github.com/ethanboxx/seed/pull/3.

@ethanboxx I would be very glad if you help me with:

1. Debug `Tide` or at least write an issue in their repo. I had to use `SET_COOKIE` headers directly because either I don't know how to use `insert/remove_cookie` or there are bugs. See the code in the PR above.

1. Another strange `Tide` behaviour - sometimes CORS doesn't work. One fetch (on app reload or click the button) from xx fetches doesn't work (both Chrome and Firefox on Windows). I don't know why and it's pretty random.

1. I've temporary commented out `secure` flag for cookies in that PR. It doesn't work in Chrome on HTTP `localhost`. Firefox probably has an exception for `localhost`. I don't know how to resolve it properly - write a backend exception? Force to run both client and server on the HTTPS somehow?

MartinKavik:
> CSRF is avoided by the line.
> `allow_origin(Origin::from(CLIENT))`
> Which instructs browsers to not send requests from malicious sites.

I was thinking about CSRF and the architecture. I found some possible problems:

A. You have to remember to update `CLIENT` when you change the real client domain. I'm pretty sure that I would forget to do that when I, for instance, change the app name in Netlify GUI - a few clicks and all my app clients can't make any successful requests anymore. It would be even more painful once you decide to scale your server horizontally.

B. What happens when you need to add more clients? It will be pretty annoying. And you probably can't trust to all of them so it's possible your whitelisted domain will try to perform CSRF attack on the other domain.

C. You can't log out without internet connection.

---

Possible solution

1. Server returns a `confirmation_token` in the response from `/sign-in` or `/signed-in` if the user is logged in. Its value would be a a complex unique random value. We can also use the `jti` (JWT Id) value if there are no security problems.

1. The same `confirmation_token` value is also written in the JWT (as a custom field like `cft`) or we just use `jti` to reduce JWT size (if it isn't a problem, again).

1. Client saves `confirmation_token` to a safe location in memory.

1. Client sends a header `confirmation_token` with the token's value with each request (including GET requests?).

1. Server verifies each request by comparing `confirmation_token` in the header and in the JWT payload extracted from `httpOnly` `secure` cookie. If there isn't a header `confirmation_token` or if the header value is invalid, then the user has been logged out or it's CSRF attack. (Server needs to verify also JWT payload with signature to make sure that it contains the correct `confirmation_token` value).

CORS - server would allow the domain of the incoming request. (The alternative to forbidden wildcard `*`.)

_Note:_ I've named it `confirmation_token` instead of `csrf_token` because it's not used only as CSRF prevention but also to log out. It _confirms_ that the request is valid and the user still wants to be logged in. 

---

Extra

- I. Perhaps we should mention symmetric vs asymmetric encryption or even implement asymmetric one.
- II. We can include the link to the [video about JWT](https://www.youtube.com/watch?v=67mezK3NzpU) in the example README.
- III. Perhaps we should set cookie's and JWT's expiration time to the same value.

---

Do you see some security holes or any other problems?

flosse:
@eopb would you mind to update this PR?

eopb:
@flosse I'm afraid that it has been too long since I started this PR for me to feel comfortable revisiting it. Thank you @MartinKavik for all of the awesome feedback, I learned a lot. Sorry I wasn't able to see this example to completion. I am going to close this now.

flosse:
No problem, thanks for your reply and clarification!

PR #509, opened 2020-07-25
Few makefile changes/cleanups

sagiegurari:
Pretty amazing project. Figured I could help make the build file a bit more clean and shorter.

* Remove all workspace=false and instead defined default_to_workspace=false in the config
* Make all core cargo-make tasks private with default namespace and remove clear=true from all seed tasks
* Remove installation instructions and instead depend on core cargo-make installation tasks
* Replace rust for_each implementation with duckscript which is much shorter, simpler and faster (in case you don't have cargo-script installed).
* Remove direct call to cargo script and instead use script_runner rust with rust file path - wasn't able to test this one properly.

while testing i wasn't able to properly test:
populate_styles - that script doesn't seem to work for me. i'm not sure why.
geko invocation - wasm tried to invoke it but i didn't have it available.
all dev and testing was done in gitpod env so I could only use whatever was preinstalled.

question about workspace. I saw its false for each task so instead I set it globally in the config.
so no need to define it for each task. but... is this crate even setup as a workspace? why was it needed to begin with?


sagiegurari:
Migrated populate_styles as well into the makefile and converted to duckscript.
this allows for faster builds
1. no need to install cargo-script first time which takes time
2. no need to compile the populate_styles.rs file which again, takes time.

this does require to reinstall the latest cargo-make version with --force so if needed i'm ok with
1. bumping cargo make version
2. setting a new min cargo-make version in the makefile

MartinKavik:
Awesome job! I was pretty excited when I saw the notification that `cargo-make` master improved our scripts 🎉 I remember that I had tendency to wash my hands when I've finished to write those old Rust scripts.., it's a very nice improvement.

I've tested your changes on my machine (Windows) and everything seems to work without any problems.

Please update CHANGELOG and rebase onto `master` and I'll merge it.

---

>while testing i wasn't able to properly test ...
... is this crate even setup as a workspace? why was it needed to begin with?

Are these statements / questions still relevant or have you already found answers during the work on the latest commits? 

---

P.S. Sorry for the late response, I was on holiday and saw the notification only thanks to a hill where was a strong enough signal for internet connection 😃

sagiegurari:
Thanks a lot for the review and comments. holidays are more important than PRs 😃 
This project really increased cargo-make visibility so i'm super happy to help where i can.
I'll update the changelog as requested.

that comment was a question since i had issues testing few things and started to dig into the structure of this project which confused me a bit. but if you did some regression testing as well, than we are fine.
if later on we need to fix anything, feel free to mention me and i'll help out.

MartinKavik:
> This project really increased cargo-make visibility

I recommend it wherever I go, I'm happy it helps 🙂 It's quite possible I'll integrate it also into [Seeder](https://github.com/MartinKavik/seeder). And ideally once Seeder is ready and Hooks/Style are integrated into Seed I want to focus more on marketing so we get more users and sponsors to make Seed a mainstream framework - I hope it'll make `cargo-make` also more popular.

> the structure of this project which confused me a bit

When you find weird/confusing project parts, scripts, APIs, etc., just create a new issue or write on the chat, every feedback is very welcome!

> if later on we need to fix anything, feel free to mention me and i'll help out.

Thanks! And let me know when you want to try/test some new `cargo-make` / `duckscript` features 🙂 

---

Changelog looks good and CI pipeline is ok, merging.

MartinKavik:
@sagiegurari 
I've found an issue probably related to the changes in this PR: `cargo make verify` passes even if there are examples with compilation errors. See task `Verify` in https://github.com/seed-rs/seed/pull/467/checks?check_run_id=985053859 - it shows errors but it doesn't kill verification.
My first tip is that `exec --fail-on-error cargo make --cwd ${example_root} ${task}` isn't working as expected. Could you look at it please? 

cc @TatriX

PR #513, opened 2020-08-09
Annotate fetch::Response with derive Debug

Ben-PH:
the underlying web_sys::Response implements debug, so may as well take advantage of it.

### DIFF
--- a/src/browser/fetch/response.rs
+++ b/src/browser/fetch/response.rs
@@ -8,6 +8,7 @@ use wasm_bindgen_futures::JsFuture;
 /// To get one you need to use [`fetch`](./fn.fetch.html) function.
 ///
 /// [MDN reference](https://developer.mozilla.org/en-US/docs/Web/API/Response)
+#[derive(Debug)]
 pub struct Response {
     pub(crate) raw_response: web_sys::Response,
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,7 @@
 - Added `WebSocket` + related items (#8).
 - Exposed `App::mailbox`.
 - Added `streams::backoff` + updated `websocket` example.
+- Added `#[derive(Debug)]` to `fetch::Response`
 
 ## v0.6.0
 - Implemented `UpdateEl` for `Filter` and `FilterMap`.


src/browser/fetch/response.rs
```diff
@@ -8,6 +8,7 @@ use wasm_bindgen_futures::JsFuture;
 /// To get one you need to use [`fetch`](./fn.fetch.html) function.
 ///
 /// [MDN reference](https://developer.mozilla.org/en-US/docs/Web/API/Response)
+#[derive(Debug)]
```
MartinKavik, line 11: Can you use `log!(response.raw_response())` instead? Rust doesn't implement `Debug` automatically for all items because it would break encapsulation in some cases - i.e. there are cases that you don't want to allow to read the content of private properties. And our `Response` contains only one private property so I'm not sure if we want to implement `Debug` for it. Another reason against it is that native JS/`web_sys` logs are often pretty useless / messy.

Why you need it?
How output of `log!(response.raw_response())` looks like now? 

And if we decide to implement `Debug`, perhaps the custom implementation would be more useful.

Ben-PH: I'd like to have it because some of my messages have `fetch::Response` as an assosciated type. In the dev process, my `update`s `match` statement has `_ => log!("impl me", msg)`, which. I can't `#derive(Debug)]` on a `Message` this PR allows for `Messages` with the assosciated type to fit in with that work-flow.

at the moment, `log!(resp.raw_response())` doesn't put out anything ugly, but I don't have too many use-cases: Network error and 401 Unauthorized at the moment.

```
NetworkError(
    JsValue(TypeError: NetworkError when attempting to fetch resource.
    ),
)

Response {
    obj: Object {
        obj: JsValue(Response),
    },
}
```

I'm happy for this PR to be rejected on the basis of hand-rolling an `impl Debug`. I can use a local edit with this change in it while I use the work-flow.

MartinKavik: I can imagine that custom `Debug` for `Response` should return something like:
```rust
Response {
    url: [url]
    status: [number + status text?],
    headers: [output of a new method Response::headers()?],
    body: [output of Response::text()?],
}
```
What do you think?

Ben-PH: I haven't come across any messy `Response`s to understand how this would/would not be an improvement. It is likely to be an improvement over the `derive`d output based on my current experience.

I could see there being an issue with printing out the text of a large body. If I understand it correctly, you're trying to, e.g., fetch a large file. If JS has the ability to put it up in the console as an object, or just prints `<large output. Table of requests in the Developer tools>`?.

Alternatively, similar to how it puts XHR objects into the console on response error, it puts the entire response into the console. Not sure if or how that can be done, though :/

MartinKavik: Ok, so I suggest these steps:
1) Create a new issue "Improve Response Debug output" in this repo and add reference to this PR or ideally to this discussion into it.
2) Merge this PR with `#[derive(Debug)]` (and with resolved previous review comment about Seed version/changelog)
3) You can try to write a new PR that would resolve the new issue or I'll try it in the future.

Ben-PH: Done: #514

Issue #535, opened 2020-09-07
Try to remove #[wasm_bindgen(start)]


MartinKavik:
See https://github.com/thedodd/trunk/issues/19

MartinKavik:
UPDATE: It seems that `wasm-pack` doesn't support `fn main` yet and it's still not recommended to use `wasm-bindgen` directly because you would have to manage `wasm-bindgen` and `wasm-bindgen-cli` installations to match the exact `wasm-bindgen` versions in `Cargo.lock`.

PR #561, opened 2020-12-02
Have a working example with trunk

seanybaggins:
Closes issue #535.

Created an example that uses trunk instead of the `cargo make` setup.

Some questions...
Is it worth making or replacing examples to use trunk?

Things that should still be done...
Update documentation for users to use trunk if that is a direction that is chosen.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 **.wasm
 **/*.rs.bk
 **/*.log
+dist
 
 # --- Dependencies and Build artifacts ---
 **/node_modules
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -117,6 +117,7 @@ features = [
 [workspace]
 members = [
     "examples/animation",
+    "examples/animation-no-wasm-bindgen-start",
     "examples/auth",
     "examples/bunnies",
     "examples/component_builder",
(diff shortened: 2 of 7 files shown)


MartinKavik:
@seanybaggins Thankd for the PR!

> Is it worth making or replacing examples to use trunk?

- Yeah, everything should be slowly migrated to use Trunk (including docs and quickstart(s)).

I would just modify the original examples and remove as much code as possible. Less code => less confusion, less bugs, better maintainability.

But there are three main blockers for Trunk integration:
1. `wasm-opt` integration - https://github.com/thedodd/trunk/issues/5#issuecomment-682152671
1. Clean previous build artifacts - https://github.com/thedodd/trunk/pull/69
1. Resolve infinite watch & rebuild loop - https://github.com/thedodd/trunk/pull/84

We can live without automatic `wasm-bindgen` installation, auto-reloading etc., but these three things reduce developer experience and app quality significantly and I can't recommend to use it for production at the moment. 

I unfortunately doesn't have a capacity to resolve those problems in the near future, so we would be glad for any help. Then we can continue with this PR.
Does it make sense?

seanybaggins:
Total sense. I will see if I can work on any of the issues you mentioned. Thanks for making Seed :)

flosse:
I also like the idea of using trunk as a standard tool. Instead of creating an additional example, I would update the documentation and migrate all the examples. So I'm closing this topic now and linking it to #611.

PR #570, opened 2021-01-18
--force flag of "cargo install" is no longer needed

tshepang:


MartinKavik:
@tshepang Thanks for another docs improvement!

_Note_: We didn't want to remove `--force` for `cargo install` from our docs immediately after the `cargo install` has been changed because many Seed users use older Rust versions. But I think this changed `cargo install` behavior is now old enough to delete it from docs.  
I'll merge the PR once CI jobs are done.

tshepang:
As a note, I checked minimum supported version of seed, and it appears that we only latest stable is supported. Is that intentional?

MartinKavik:
> 
> 
> As a note, I checked minimum supported version of seed, and it appears that only latest stable is supported. Is that intentional?

Yes
- We often need some new Rust features to implement Seed features or to do a proper refactor. 
- We don't want to be too incompatible with nightly Rust nightly because some users use nighly Rust and there are some features (e.g. debugging) that works as expected only on nightly.
- Most new Rust versions break CI because they introduce new Clippy or Fmt rules.
-  etc.

PR #572, opened 2021-01-22
prevent system from intercepting links with a download attribute

blunted2night:
This PR prevents seed from intercepting links with a download attributes

```html
<a download="filename.ext" href="/files/filename.ext">filename.ext</a>
```

### DIFF
--- a/src/browser/service/routing.rs
+++ b/src/browser/service/routing.rs
@@ -87,7 +87,13 @@ pub fn setup_link_listener(notify: impl Fn(Notification) + 'static) {
         event.target()
             .and_then(|et| et.dyn_into::<web_sys::Element>().ok())
             .and_then(|el| el.closest("a[href]").ok().flatten())
-            .and_then(|href_el| href_el.get_attribute("href"))
+            .and_then(|href_el| {
+                if href_el.has_attribute("download") {
+                    None
+                } else {
+                    href_el.get_attribute("href")
+                }
+            })
             // The first character being / or empty href indicates a rel link, which is what
             // we're intercepting.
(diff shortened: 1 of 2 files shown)


src/browser/service/routing.rs
```diff
@@ -87,7 +87,13 @@ pub fn setup_link_listener(notify: impl Fn(Notification) + 'static) {
         event.target()
             .and_then(|et| et.dyn_into::<web_sys::Element>().ok())
             .and_then(|el| el.closest("a[href]").ok().flatten())
-            .and_then(|href_el| href_el.get_attribute("href"))
+            .and_then(|href_el| {
+                if href_el.get_attribute("download").is_none () {
```
MartinKavik, line 91: Please change the the line:
```suggestion
                if href_el.has_attribute("download") {
```
There are some potential [problems](https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute#non-existing_attributes) with `get_attribute` and `has_attribute` better describes our intentions and there is a chance it's slightly faster.

MartinKavik:
@blunted2night 
Please fix the Clippy errors and update CHANGELOG and I'll merge it, thanks!

blunted2night:
what should I use for the version line?

MartinKavik:
> what should I use for the version line?

If you mean the version in CHANGELOG - Just add a new item below `[unreleased]`. 

Also it would be nice to squash commits into one.

blunted2night:
Sorry, I was looking at the warp change log. Thanks.

MartinKavik (approved):

PR #575, opened 2021-02-17
Consider title trim and empty case

onevcat:
Ref: https://github.com/MartinKavik/seed-app-todomvc/pull/3


MartinKavik:
FYI Linter errors fixed and the code slightly changed in https://github.com/seed-rs/seed/pull/576

PR #578, opened 2021-02-27
feat(fetch): Add Response.blob

### LINKED ISSUE #577: Feature request: add .blob() to fetch
It would be nice to have [.blob()](https://developer.mozilla.org/en-US/docs/Web/API/Body/blob) method returning [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob).

glennsl:
Fixes #577

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Prevent link listener from intercepting links with the `download` attribute.
 - Added examples `record_screen` and `e2e_encryption`.
 - Adapted to Rust 1.50.0.
+- Added `Response::blob`
 
 ## v0.8.0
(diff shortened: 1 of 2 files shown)


MartinKavik:
Nice! Please update CHANGELOG and I'll merge it.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - Prevent link listener from intercepting links with the `download` attribute.
 - Added examples `record_screen` and `e2e_encryption`.
 - Adapted to Rust 1.50.0.
+- Added `Response::blob`
```
glennsl, line 8: Done!

PR #583, opened 2021-03-07
refactor: hide markdown functionality behind optional feature

glennsl:
Just to lighten the payload a little. I've probably missed something though, and if so just let me know.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Adapted to Rust 1.50.0.
 - Added `Response::blob`
 - Added `panic-hook` feature, enabled by default, to conditionally include `console_error_panic_hook`
+- [BREAKING] Hid markdown funnctionality behind optional `markdown` feature
 
 ## v0.8.0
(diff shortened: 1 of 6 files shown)


MartinKavik:
Do you have some numbers? E.g. how much is compilation faster or the wasm file size smaller?

glennsl:
It doesn't actually impact the wasm size. For some reason I thought it would. It does however reduce compile time by about 10% for simple examples.

Personally I'd prefer to keep a stronger separation so as not to accidentally trigger the inclusion of another dependency and a significant increase in code size. But I also try to keep dependencies few in general, to keep things simple and reduce the number of things that change over time, and can therefore go wrong, both at build-time and run-time. Adding another feature certainly doesn't make it simpler though, so there's definitely a trade-off there.

MartinKavik:
> But I also try to keep dependencies few in general, to keep things simple and reduce the number of things that change over time, and can therefore go wrong

I agree 100%. I'll merge the PR.

Btw I'll try to write a Rust frontend app almost without dependencies and with `#[no_std]` in [MoonZoon](http://moonzoon.rs/). I'm curious how far I'll be able to go.

Issue #584, opened 2021-03-11
Would it be possible to accept event handlers that do not implement `Clone`?


glennsl:
One use case for this is having a component with an unknown number of "actions", effectively a `Vec<Box<dyn FnOnce() -> Msg>>` or alternatively just `Vec<Msg>`.

Unfortunately, since `ev` requires the event handler implements `Clone`, the vec of closures is not possible because trait objects can not include non-auto-traits like `Clone`, i.e. `Vec<Box<dyn FnOnce() -> Msg + Clone>>` is not accepted by the compiler. And `Vec<Msg>` requires `Ms` to implement `Clone`, which is infectious and infeasible for all but the simplest of `Msg` types.

I have made some attempts to implement various alternative versions of `ev`, but have so far been unable to for various reasons related to moving and borrowing. I haven't even been able to make a simpler version without the macro and only accepting a handler that returns `Ms`. It seems to me that this should be doable though.

MartinKavik:
Could you provide a simple example / use-case? I'm not sure I understand your motivation.

However I'm also not sure if it's possible - event handlers have to be callable multiple times - e.g. user clicks multiple times on a button. It means event handlers have to implement `Fn` ([related code](https://github.com/seed-rs/seed/blob/89245248e3d129967725361130437a5e0affef4b/src/virtual_dom/event_handler_manager/event_handler.rs#L8) in Seed). Function `ev` [accepts](https://github.com/seed-rs/seed/blob/28498c0f2dfd3f9a1d59535c385b856a62854805/src/browser/dom/event_handler.rs#L146) `FnOnce` + `Clone` because this combination can be "transformed" to `Fn`.

glennsl:
I just discovered that the `component_builder` example does something similar to what I want, but with the requirement that `Msg` implements `Clone`. It uses `Vec<Rc<dyn Fn(MouseEvent) -> Ms>>`. So basically that, but without needing `Clone` on `Msg`.

Another use case is optional event handlers, where the type of the event handler would have to be explicitly annotated because of course it can't be inferred if it's not specified. A trait object or `Msg` value passed in directly would solve that as well I think.

The reason I'd like to avoid `Clone` on `Msg` is that it requires everything referenced by it to also implement `Clone`. Which isn't possible to derive for external types, for example.

[...]

PR #586, opened 2021-03-13
refactor(examples/component_builder): declone Msg and simplify click handler

glennsl:
As suggested in #584.

`Clone` is actually not needed on `Msg` because of the event handler, but because clippy is being a bit anal. So to reduce confusion I've removed the derive and disabled the clippy warning instead. (edit: also done for all other examples now)

I've also simplified the event handler a bit by removing the `MouseEvent` argument, partly because I think it's good design in general to not expose unnecessary implementation details, but also partly because it illustrates that `Msg`s must be passed in as a closure even if it doesn't take any arguments, unless it implements `Clone`.

### DIFF
--- a/examples/animation/src/lib.rs
+++ b/examples/animation/src/lib.rs
@@ -68,12 +68,12 @@ impl Default for Car {
 //    Update
 // ------ ------
 
-#[derive(Copy, Clone)]
 enum Msg {
     Rendered(RenderInfo),
     SetViewportWidth,
 }
 
+#[allow(clippy::needless_pass_by_value)]
 fn update(msg: Msg, model: &mut Model, orders: &mut impl Orders<Msg>) {
     match msg {
         Msg::Rendered(render_info) => {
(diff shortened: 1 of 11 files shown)


MartinKavik:
> Clone is actually not needed on Msg because of the event handler, but because clippy is being a bit anal. So to reduce confusion I've removed the derive and disabled the clippy warning instead.

I also recommend in seed-rs.org docs to derive `Copy` for all items where it's possible. And I try to respect Clippy where it makes sense. However I see it's confusing for newcomers because it breaks the "no-Clone `Msg`" rule so I agree with the change.

> I've also simplified the event handler a bit by removing the MouseEvent argument, partly because I think it's good design in general to not expose unnecessary implementation details, but also partly because it illustrates that Msgs must be passed in as a closure even if it doesn't take any arguments, unless it implements Clone.

Yeah, good idea. If I remember correctly, my motivation was to show how to pass events if somebody needs more details or call something like `event.stop_propagation()`. However it's not important in this example I think.

Thanks!

PR #588, opened 2021-03-20
add example of using CustomEvent to custom_elements example

glennsl:
This is rather contrived, but does an OK job at illustrating how to use custom events I think. I'm very open to suggestions though.

It might also be an idea to convert this into a component builder for a nicer API, but maybe that will complicate the example too much.


glennsl:
`cargo make verify` passes locally, not sure why it fails on CI with:

```
 Error during execution of `cargo metadata`:     Updating crates.io index
error: failed to select a version for the requirement `shipyard = "^0.4.1"`
candidate versions found which didn't match: 0.5.0
```

`cargo update` locally yields the same error though

MartinKavik:
Please remove `**/Cargo.lock` from `.gitignore` and push all your locks to the repo. I started to do it in all my repos because I have to often resolve similar dependency issues.
Code looks good on the first glance, I'll merge it then.

glennsl:
Done. Yea we do the same.

PR #600, opened 2021-05-17
Add `At::Role`

fosskers:
This PR adds a `Role` variant for the giant `At` enum, so that the `role` attribute can be used like so:

```html
<nav role="navigation"></nav>
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## [unreleased]
+
 - Added helpers for wheel event: `wheel_ev` and `to_wheel_event`.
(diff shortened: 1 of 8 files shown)


MartinKavik:
please run `cargo make verify` on your local machine

fosskers:
Ah ok, I was just doing normal `cargo clippy`.

MartinKavik:
It's written in `CONTRIBUTING.md`, but I'm not sure how much GitHub forces to read it by new contributors (if at all).

fosskers:
I'm sure I read it when I first found seed, but since I haven't contributed directly to the project until now, I must have forgotten. Either way I'll fix this up today.

PR #613, opened 2021-06-24
Intersection Observer example

allan2:
Martin, I'm very grateful for your help in making this example.

I changed the threshold to be a single value instead of array because the array wasn't working. `is_intersecting` was not updating. I did open https://github.com/rustwasm/wasm-bindgen/issues/2601 for that.

Other blocker is I don't have a `Makefile.toml`. Will I have to create one to be accepted?


MartinKavik:
> I changed the threshold to be a single value instead of array because the array wasn't working.

I don't push clearly broken commits. It works on Chrome and Firefox on Windows. The most problematic browser is often Safari or all browsers on iPhones. What's the problem? Also if it doesn't work I would assume it's a browser bug unrelated to `wasm-bindgen`.

> Other blocker is I don't have a Makefile.toml. Will I have to create one to be accepted?

Yes and add the example to the root `Cargo.toml` as a workspace member, please. We want to run it with `cargo make start intersection_observer` from the repo root, we don't want an extra `target` folder, we want it to be tested automatically by CI and for consistency. However I have no problem with the gradual `trunk` integration.

I'll do code review. Then please update the code as/if needed, squash commits and rebase on `master` so I can merge it. Thanks!

examples/intersection_observer/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+
+[lib]
+crate-type = ["cdylib"]
+
+[dev-dependencies]
+wasm-bindgen-test = "0.3.18"
+
+[dependencies]
+seed = { path = "../../" }
+js-sys = { version = "0.3.51" }
```
MartinKavik, line 15:
```suggestion
```
We don't need it - https://github.com/seed-rs/seed/blob/master/src/lib.rs#L206

examples/intersection_observer/README.md
```diff
@@ -0,0 +1,11 @@
+# Intersection Observer Example
+
+This example uses the [Intersection Observer V1 API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) through web-sys.
```
MartinKavik, line 3:
```suggestion
This example uses the [Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) through `web-sys`.
```
Why `V1`?

examples/intersection_observer/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+The intersection observer is created. We tell it to watch the red box by giving the element reference to the observer.
+When the red box is entirely in view, the label "Is intersecting entirely" will display a value of `true`.
```
MartinKavik, line 6: What do you think about something like _"Is completely visible"_  instead of _"Is intersecting entirely"_?

examples/intersection_observer/src/lib.rs
```diff
@@ -0,0 +1,124 @@
...
+
+            // ---- observer options ----
+            let mut options = IntersectionObserverInit::new();
+            options.threshold(&JsValue::from(1));
+            // ---- observer ----
+            let observer =
+                IntersectionObserver::new_with_options(callback.as_ref().unchecked_ref(), &options)
+                    .unwrap();
+
+            observer.observe(&model.red_box.get().unwrap());
```
MartinKavik, line 55: You don't see it here on GitHub, but it's badly formatted.
image
Please run `cargo fmt`. Another reason why we need `Makefile.toml` - we want to automatically lint the example by CI.

examples/intersection_observer/src/lib.rs
```diff
@@ -0,0 +1,124 @@
...
+fn view_info(observer_entries: Option<&Vec<IntersectionObserverEntry>>) -> Option<Node<Msg>> {
+    let entry = observer_entries?.first()?;
+    Some(div![
+        style! {
+            St::Position => "fixed",
+            St::Background => "white",
+        },
+        div![format!("Target: {:#?}", entry.target().id()),],
+        div![format!(
+            "Is intersecting entirely: {:#?}",
```
MartinKavik, line 80:
```suggestion
        div![format!("Target: {}", entry.target().id()),],
        div![format!(
            "Is intersecting entirely: {}",
```
I think `ToString` is enough.

examples/intersection_observer/index.html
```diff
@@ -0,0 +1,20 @@
...
+</head>
+
+<body>
+    <div id="app"></div>
+    <script type="module">
+        import init from '/pkg/package.js';
+        init('/pkg/package_bg.wasm');
+    </script>
+</body>
+</html>
```
MartinKavik, line 20: image

MartinKavik (changes requested):

examples/intersection_observer/README.md
allan2: I removed it. I wanted to distinguish between the unofficial V2 https://caniuse.com/intersectionobserver-v2

examples/intersection_observer/README.md
allan2: Good idea!

allan2:
Superseded by PR #615

PR #617, opened 2021-07-22
Return response with details from the server on error

### LINKED ISSUE #616: Return error details in `FetchError`
I was using the `fetch()` API as shown in the [example](https://github.com/seed-rs/seed/blob/480141ce9e520c07e60ddae58244edb40c9f55e9/examples/server_integration/client/src/example_a.rs), using `check_status()` helper function. Unfortunately, this function has a pretty significant flaw - it hides the error details returned from the backend.

Let's say, backend returns error `500` with following response:  
```json
{"details":"cannot connect to the database"}
```

If I'll `log!()` such error in Seed, I'll get this:  
```
StatusError(
    Status {
        code: 500,
        text: "Internal Server Error",
        category: ServerError,
    },
)
```

There's no way to figure out what happened in the server, as the `details` are now gone. It would be great to have `response` in `Status` that would contain the server response. 

For now, I'm not using `check_status()` and manually check for `status()` to return an error, but it feels a bit clunky. Is there any better way?

aklajnert:
Resolves #616

This PR introduces a new enum variant for `FetchError` that contains the response status together with the JSON response returned from the server.

Initially, I wanted to change the `check_status()` method to try to return the new variant first, and in case of JSON parse error just return the regular `StatusError`, but it would require changing the method to `async` which would be a breaking change. Therefore I decided to add another method.

Here's how it looks when you `log!()` such error:  
```
DetailedStatusError(
    Status {
        code: 500,
        text: "Internal Server Error",
        category: ServerError,
    },
    Object({
        "details": String("cannot connect to the database",),
    }),
)
```

### DIFF
--- a/src/browser/fetch/response.rs
+++ b/src/browser/fetch/response.rs
@@ -101,6 +101,26 @@ impl Response {
         }
     }
 
+    /// Check that response status is ok (2xx), in case of error return a tuple of error and Option<String>
+    /// with the error response from a server.
+    ///
+    /// ```rust
+    /// fetch(url).await?.check_detailed_status()?
+    /// ```
+    ///
+    /// # Errors
+    /// Returns a tuple of `FetchError` and `Option<String>` with error details if status isn't 2xx.
+    pub async fn check_detailed_status(
+        self,
+    ) -> std::result::Result<Self, (FetchError, Option<String>)> {
+        let status = self.status();
+        if status.is_ok() {
+            Ok(self)
+        } else {
+            Err((FetchError::StatusError(status), self.text().await.ok()))
+        }
+    }
+
     /// Get underlying `web_sys::Response`.
     ///
     /// This is an escape path if current API can't handle your needs.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 - Removed internal `serde_json` usage in favour of `serde-wasm-bindgen`. This reduces final binary size for downstream users.
 - [BREAKING] Removed the deprecated `browser::service::fetch` module.
 - Element macros like `div!` can now contain `Iterator`s inside of `Option` values. Previously only one or the other was possible.
+- Add method to return detailed error response from server with `FetchError`.
 
 ## v0.8.0
 


glennsl:
This seems a bit bolted-on TBH. Adding another case to`FetchError` means you'll also have to check for `DetailedStatusError` wherever you check for `StatusError`, which is both boilerplate-y and rather error-prone. I would also consider this a breaking change, since consumers who take advantage of exhaustiveness checking of enums will have to change their code to include this new case.

What about having `check_detailed_status` return a tuple instead of adding a new case?

Also, why limit this to only work with JSON responses? What about XML or whatever newfangled format that will replace JSON? Could it not return the response object instead, and have the consumer decide how to decode the body? Or, alternatively, pass a decoder in as an argument?

aklajnert:
Thanks @glennsl, these were all fair points that I've missed. I didn't really consider that adding an enum variant is a breaking change, but you're right. Also, I was a bit selfish and didn't consider other response formats, as I work mostly with JSON.  
I've changed my code to fix that.

src/browser/fetch/response.rs
```diff
@@ -92,6 +92,35 @@ impl Response {
...
+    ) -> std::result::Result<Self, (FetchError, Option<T>)>
+    where
+        T: std::str::FromStr,
+    {
+        let status = self.status();
+        if status.is_ok() {
+            Ok(self)
+        } else {
+            let value = if let Ok(text) = self.text().await {
+                T::from_str(&text).ok()
```
flosse, line 116: How about returning the raw text and leaving it up to the user to decide how to process the data?

aklajnert: Huh, that's a very fair point. It'll simplify the code and make it more useful (as the potential `ParseError` from `from_str()` won't be lost). I'll change it.

flosse:
BTW: This is my current solution that works for me:

```rust
#[derive(Debug)]
pub enum Error<E> {
    Fetch(FetchError),
    Api(boundary::Error<E>),
}

impl<E> From<FetchError> for Error<E> {
    fn from(e: FetchError) -> Self {
        Self::Fetch(e)
    }
}

pub type Result<T, E> = result::Result<T, Error<E>>;

async fn to_result<T, E>(res: Response) -> Result<T, E>
where
    T: for<'de> Deserialize<'de> + 'static,
    E: for<'de> Deserialize<'de> + 'static,
{
    if res.status().is_ok() {
        let data = res.json().await?;
        Ok(data)
    } else {
        let error = res.json().await?;
        Err(Error::Api(error))
    }
}
```
And this is how I use it:

```rust
pub async fn get_json<T, E>(url: &str) -> Result<T, E>
where
    T: for<'de> Deserialize<'de> + 'static,
    E: for<'de> Deserialize<'de> + 'static,
{
    let res = fetch(url).await?;
    to_result(res).await
}

pub async fn post_json<R, T, E>(url: &str, req: &R) -> Result<T, E>
where
    R: Serialize,
    T: for<'de> Deserialize<'de> + 'static,
    E: for<'de> Deserialize<'de> + 'static,
{
    let req = Request::new(url).method(Method::Post).json(req)?;
    let res = fetch(req).await?;
    to_result(res).await
}
```

glennsl (approved):
Having it return `String` means it can't be used for binary formats like BSON or protobuf, which might be used for structured errors. Still I'd expect this to cover 95% of use cases, it's not a breaking change and it doesn't interfere with anything else, so looks good to me!

MartinKavik:
Thanks for PR!

Also thanks @flosse for comments and @glennsl for excellent code review / suggestions.

Let's wait a bit until https://github.com/seed-rs/seed/pull/618 is merged, then we can rebase (to also resolve clippy/compiler errors (?)) and merge it.

aklajnert:
@MartinKavik - done

PR #618, opened 2021-08-05
Replace `serde_json`

fosskers:
This PR replaces `serde_json` with [serde-wasm-bindgen](https://lib.rs/crates/serde-wasm-bindgen) in order to reduce dependency load for downstream users of Seed.


Cargo.toml
```diff
@@ -33,8 +33,8 @@ rand = { version = "0.8.0", features = ["small_rng"] }
 # https://docs.rs/getrandom/0.2.0/getrandom/#webassembly-support
 getrandom = { version = "0.2", features = ["js"] }
 serde = { version = "1.0.117", features = ['derive'] }
-serde_json = "1.0.59"
-wasm-bindgen = { version = "0.2.70", features = ["serde-serialize"] }
```
fosskers, line 37: Ensures that `wasm-bindgen` doesn't transitively depend on `serde_json`.

src/browser/fetch/form_data.rs
```diff
@@ -60,9 +61,12 @@ impl FormData {
...
     pub fn append_json<T>(&mut self, name: &str, data: &T) -> Result<()>
     where
-        T: Serialize + ?Sized,
+        T: Serialize,
     {
-        let str = serde_json::to_string(data).map_err(FetchError::SerdeError)?;
+        // FIXME Can a different `append` be used to append a `JsValue` directly?
+        let str = swb::to_value(data)?
+            .as_string()
+            .ok_or(FetchError::ConversionError)?;
```
fosskers, line 69: :-1:  for efficiency since whole JSON structures cannot be added to the `FormData`.

src/browser/fetch/request.rs
```diff
@@ -88,9 +89,12 @@ impl<'a> Request<'a> {
...
     /// return `FetchError::SerdeError`.
-    pub fn json<T: Serialize + ?Sized>(mut self, data: &T) -> Result<Self> {
-        let body = serde_json::to_string(data).map_err(FetchError::SerdeError)?;
-        self.body = Some(Cow::Owned(body.into()));
+    pub fn json<T>(mut self, data: &T) -> Result<Self>
+    where
+        T: Serialize,
+    {
+        let body = swb::to_value(data)?;
+        self.body = Some(Cow::Owned(body));
```
fosskers, line 97: :+1: for efficiency since we don't need to allocate an intermediate `String`.

src/browser/fetch/response.rs
```diff
@@ -35,8 +37,15 @@ impl Response {
...
-        serde_json::from_str(&text).map_err(FetchError::SerdeError)
+        let js: JsValue = self
+            .raw_response
+            .json()
+            .map_err(FetchError::PromiseError)
+            .map(JsFuture::from)?
+            .await
+            .map_err(FetchError::PromiseError)?;
+
+        Ok(swb::from_value(js)?)
```
fosskers, line 48: :+1:  for efficiency since we don't need to allocate an intermediate `String`.

src/browser/service/fetch.rs
```diff
@@ -343,10 +344,8 @@ impl Request {
     /// Serialize a Rust data structure as JSON; eg the payload in a POST request.
     /// _Note_: If you want to setup `Content-Type` header automatically, use method `send_json`.
     pub fn body_json<T: Serialize>(self, body_json: &T) -> Self {
-        let json =
-            serde_json::to_string(body_json).expect("fetch: serialize body to JSON - failed");
-        let json_as_js_value = JsValue::from_str(&json);
-        self.body(json_as_js_value)
+        let json = swb::to_value(body_json).expect("fetch: serialize body to JSON - failed");
+        self.body(json)
```
fosskers, line 348: :+1: , no intermediate String.

src/browser/service/fetch.rs
```diff
@@ -550,42 +549,57 @@ impl Request {
...
                                 }),
                             },
-                            Ok(value) => FetchObject::<T> {
-                                request,
-                                result: Ok(ResponseWithDataResult {
-                                    raw: response.raw,
-                                    status: response.status,
-                                    data: Ok(value),
-                                }),
+                            Ok(js_value) => match swb::from_value(js_value.clone()) {
```
fosskers, line 583: :+1: , no intermediate String.

src/browser/service/routing.rs
```diff
@@ -3,18 +3,18 @@ use super::super::{
...
+/// Add a new route using history's `push_state` method.
+///
+/// # References
+/// * [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/History_API)
 pub fn push_route<U: Into<Url>>(url: U) -> Url {
     let url = url.into();
     // We use data to evaluate the path instead of the path displayed in the url.
-    let data =
-        JsValue::from_str(&serde_json::to_string(&url).expect("Problem serializing route data"));
+    let data = swb::to_value(&url).expect("Problem serializing route data");
```
fosskers, line 17: :+1: , no intermediate String.

src/browser/service/routing.rs
```diff
@@ -32,12 +32,10 @@ pub fn setup_popstate_listener(
             .dyn_ref::<web_sys::PopStateEvent>()
             .expect("Problem casting as Popstate event");
 
-        let url = match ev.state().as_string() {
-            Some(state_str) => {
-                serde_json::from_str(&state_str).expect("Problem deserializing popstate state")
-            }
+        let url = match swb::from_value(ev.state()) {
```
fosskers, line 35: :+1: , no intermediate String.

src/browser/url.rs
```diff
@@ -40,9 +41,7 @@ impl Url {
     /// * [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/History_API)
     pub fn go_and_push(&self) {
         // We use data to evaluate the path instead of the path displayed in the url.
-        let data = JsValue::from_str(
-            &serde_json::to_string(&self).expect("Problem serializing route data"),
-        );
+        let data: JsValue = swb::to_value(&self).expect("Problem serializing route data");
```
fosskers, line 44: :+1: , no intermediate String.

src/browser/url.rs
```diff
@@ -57,9 +56,7 @@ impl Url {
     /// * [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/History_API)
     pub fn go_and_replace(&self) {
         // We use data to evaluate the path instead of the path displayed in the url.
-        let data = JsValue::from_str(
-            &serde_json::to_string(&self).expect("Problem serializing route data"),
-        );
+        let data: JsValue = swb::to_value(&self).expect("Problem serializing route data");
```
fosskers, line 59: :+1: , no intermediate String.

src/browser/web_socket.rs
```diff
@@ -117,8 +128,13 @@ impl WebSocket {
...
     /// Returns error when JSON serialization or sending fails.
-    pub fn send_json<T: Serialize + ?Sized>(&self, data: &T) -> Result<()> {
-        let data = serde_json::to_string(data).map_err(WebSocketError::SerdeError)?;
+    pub fn send_json<T>(&self, data: &T) -> Result<()>
+    where
+        T: Serialize,
+    {
+        let data = swb::to_value(data)?
+            .as_string()
+            .ok_or(WebSocketError::ConversionError)?;
```
fosskers, line 137: :-1: , can't send whole JSON over a socket.

src/browser/web_socket/message.rs
```diff
@@ -27,11 +28,12 @@ impl WebSocketMessage {
...
-    /// - `WebSocketError::TextError` if data isn't a valid utf-8 string.
     /// - `WebSocketError::SerdeError` when JSON decoding fails.
-    pub fn json<T: DeserializeOwned + 'static>(&self) -> Result<T> {
-        let text = self.text()?;
-        serde_json::from_str(&text).map_err(WebSocketError::SerdeError)
+    pub fn json<T>(&self) -> Result<T>
+    where
+        T: DeserializeOwned + 'static,
+    {
+        swb::from_value(self.data.clone()).map_err(WebSocketError::SerdeError)
```
fosskers, line 36: :+1: , no intermediate String.

src/browser/web_storage.rs
```diff
@@ -155,13 +165,17 @@ pub trait WebStorage {
...
+        T: DeserializeOwned,
+    {
+        let item: String = Self::storage()?
             .get_item(key.as_ref())
             .map_err(WebStorageError::GetError)?
-            .ok_or(WebStorageError::KeyNotFoundError)
-            .map(|value| serde_json::from_str(&value))?
-            .map_err(WebStorageError::SerdeError)
+            .ok_or(WebStorageError::KeyNotFoundError)?;
+        let js: JsValue = JSON::parse(&item).map_err(WebStorageError::ParseError)?;
```
fosskers, line 176: :-1: , extra parsing done on the JS side to get a `JsValue`.

src/browser/web_storage.rs
```diff
@@ -174,8 +188,13 @@ pub trait WebStorage {
...
     /// [MDN reference](https://developer.mozilla.org/en-US/docs/Web/API/Storage/setItem)
-    fn insert<T: Serialize + ?Sized>(key: impl AsRef<str>, value: &T) -> Result<()> {
-        let value = serde_json::to_string(value).map_err(WebStorageError::SerdeError)?;
+    fn insert<T>(key: impl AsRef<str>, value: &T) -> Result<()>
+    where
+        T: Serialize,
+    {
+        let value = swb::to_value(value)?
+            .as_string()
+            .ok_or(WebStorageError::ConversionError)?;
```
fosskers, line 197: :-1: , can only store `String`s into this kind of storage.

fosskers (commented):
Over the 13 replacement sites, **four** saw a decrease in efficiency and **nine** saw an increase. An improvement overall.

fosskers:
The remaining clippy errors seems to be spurious, I'm not sure what to do about them.

fosskers:
Some tests are failing, but at least everything compiles now.

fosskers:
Something else might still be broken, because I'm getting weird `Reflect.get` errors that I've never seen before on code I know used to work.

fosskers:
The `Reflect.get` errors were my own fault! It's all good.

examples/service_worker/src/lib.rs
```diff
@@ -273,12 +274,8 @@ fn update(msg: Msg, model: &mut Model, orders: &mut impl Orders<Msg>) {
                 .expect("get `PushManager`");
             // @TODO: Replace the call to `subscribe` with the web_sys equivalent when wasm_bindgen releases a version >= 0.2.68.
             orders.perform_cmd(async move {
-                Msg::Subscribed(
-                    subscribe(&push_manager, key)
-                        .await
-                        .into_serde()
-                        .map_err(ServiceWorkerError::SerdeJson),
-                )
+                let js = subscribe(&push_manager, key).await;
```
MartinKavik, line 277: Optional: See the `TODO` two lines above. Then we can remove the function `subscribe`.

src/browser/fetch/form_data.rs
```diff
@@ -60,9 +61,12 @@ impl FormData {
     #[allow(clippy::missing_panics_doc)]
     pub fn append_json<T>(&mut self, name: &str, data: &T) -> Result<()>
     where
-        T: Serialize + ?Sized,
+        T: Serialize,
     {
-        let str = serde_json::to_string(data).map_err(FetchError::SerdeError)?;
+        // FIXME Can a different `append` be used to append a `JsValue` directly?
```
MartinKavik, line 66: Could you rewrite all `FIXME`s to `@TODO` for consistency so we can search through all todos/fixmes with one term?

src/browser/service/fetch.rs
```diff
@@ -550,42 +549,57 @@ impl Request {
         let fetch_result = fetch_object.result;
         let request = fetch_object.request;
 
+        // FIXME What is this giant gross match ladder?
```
MartinKavik, line 552: See the top of the file - the module is deprecated. 

There was a bunch of fetch API versions, some of them were limited by missing async support in Rust or had other problems that resulted in such monsters.

I think you can remove the file and add probably a breaking change into the changelog.

src/browser/service/routing.rs
```diff
@@ -79,8 +77,9 @@ pub fn url_request_handler(
     }
 }
 
-// Set up a listener that intercepts clicks on elements containing an Href attribute,
-// so we can prevent page refresh for internal links, and route internally.  Run this on load.
+/// Set up a listener that intercepts clicks on elements containing an Href
+/// attribute, so we can prevent page refresh for internal links, and route
+/// internally. Run this on load.
```
MartinKavik, line 82: I think there were only regular comments instead of doc comments for `setup_link_listener` and maybe `push_route` or other functions because they aren't useful for Seed users at all and should be probably marked by `pub(crate)` instead of `pub`.
Your choice how to resolve it / leave it as is.

src/browser/web_storage.rs
```diff
@@ -200,7 +223,7 @@ mod tests {
         LocalStorage::clear().unwrap();
 
         let value: String = {
-            LocalStorage::insert("a_key", "a_value").unwrap();
+            LocalStorage::insert(&"a_key", &"a_value").unwrap();
```
MartinKavik, line 226: Are references necessary?

src/browser/web_storage.rs
```diff
@@ -174,8 +191,14 @@ pub trait WebStorage {
     /// or serialize the value or insert/update the pair.
     ///
     /// [MDN reference](https://developer.mozilla.org/en-US/docs/Web/API/Storage/setItem)
-    fn insert<T: Serialize + ?Sized>(key: impl AsRef<str>, value: &T) -> Result<()> {
-        let value = serde_json::to_string(value).map_err(WebStorageError::SerdeError)?;
+    fn insert<S, T>(key: S, value: &T) -> Result<()>
+    where
+        S: AsRef<str>,
+        T: Serialize,
```
MartinKavik, line 197: Why are explicitly named typed parameters better than `impl`s? They allow to use turbofish but I don't know if it makes sense for this case.
Or I would at least replace `S` and `T` with ``K` and `V`.

src/browser/web_storage.rs
```diff
@@ -174,8 +191,14 @@ pub trait WebStorage {
...
+    {
+        let js: JsValue = swb::to_value(value)?;
+        let st: JsString = JSON::stringify(&js).map_err(WebStorageError::StringifyError)?;
+        let value: String = st.as_string().ok_or(WebStorageError::ConversionError)?;
```
MartinKavik, line 201: I would rename `js` and `st` to`value`.  All of them represent value, you don't need to use weird names and there is a chance Rust is/will be clever enough to drop old value forms as soon as possible.

MartinKavik (changes requested):
I've added some minor suggestions, then we can merge it :)

examples/service_worker/Cargo.toml
```diff
@@ -18,19 +18,20 @@ path = "src/main.rs"
...
-warp = { version = "0.2.5", optional = true }
 tracing = { version = "0.1.21", optional = true }
 tracing-subscriber = { version = "0.2.14", optional = true }
+warp = { version = "0.2.5", optional = true }
 web-push = { version = "0.7.2", optional = true }
 
 # client
 apply = { version = "0.3.0", optional = true }
 seed =  { path = "../../", optional = true }
+serde-wasm-bindgen = "0.3"
```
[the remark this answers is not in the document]
fosskers: Yeah for the backend examples it shouldn't be used.

src/browser/service/routing.rs
fosskers: I'll `pub(crate)` them.

src/browser/web_storage.rs
fosskers: https://github.com/cloudflare/serde-wasm-bindgen/pull/18 just got merged, but it's not yet released. Once released, I can add back the `?Sized` "constraint" on all the functions I had to take it off of, and then the references here (and elsewhere) can be reverted.

src/browser/web_storage.rs
fosskers: Overall I'd say `where` clauses are a bit clearer that `impl`s, but I realize that Seed uses `impl`s a lot, so if I'm just adding confusion to an established pattern, I'm okay with reverting.

Is there a difference performance-wise, or is that only when `dyn` is present?

MartinKavik: > Is there a difference performance-wise, or is that only when dyn is present?

I don't think so. I use mostly `impl` instead of "classic" generics / `where` because I don't need to come up with names and it feels a bit more natural to read to me.
The only difference I know - when you want to write a function similar to `Iterator`'s `collect` and call it with turbofish (e.g. `.collect::<Vec<_>>()`), then you have to use generics, turbofish + impl Trait causes compilation error.

`dyn` is used in Seed to simplify some types or when it's the only option. Also it may reduce a wasm file size a bit so it isn't a problem in most cases.

fosskers: Then I guess it's just our difference styles, since we both consider our own way to be easier to read :rofl: 

How about I at least change the type param names, as you suggest. Then going forward in future PRs I'll try to leave the existing style as-is.

src/browser/web_storage.rs
RReverser: FWIW you can just hide this as an implementation detail for now, either by always using `&` before calling `to_value` or by creating `Serializer` explicitly. It will be easier to update it that way in future too.

examples/service_worker/src/lib.rs
fosskers: I will leave this for follow-up work.

src/browser/web_socket.rs
```diff
@@ -118,7 +131,9 @@ impl WebSocket {
     ///
     /// Returns error when JSON serialization or sending fails.
     pub fn send_json<T: Serialize + ?Sized>(&self, data: &T) -> Result<()> {
-        let data = serde_json::to_string(data).map_err(WebSocketError::SerdeError)?;
+        let data = swb::to_value(data)?
+            .as_string()
```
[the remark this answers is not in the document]
arn-the-long-beard: Yes it is a breaking change. I got confirmation today.  https://github.com/seed-rs/seed/issues/619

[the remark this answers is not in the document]
fosskers: Okay I believe @RReverser now, I'm putting a fix together.

PR #620, opened 2021-09-03
Fix string encoding issue

### LINKED ISSUE #619: Serialize or/and Deserialize failing
Hello guys :wink: ,

Well my main project is now failling its Apis calls that contains objects to post.

### Before Update 

image

1 - From the web browser

image

2 - From the web browser

### After Update

image

3 -From the web browser

image

4 - From the web browser

image

5 - From the terminal

Here is my code 

#### Send email to backend for checking

```rust
      let email_form = FormEmail::new(model.email.clone());

            let request = Request::new("/api/email")
                .method(Method::Post)
                .json(&email_form)
                .expect("Serialization failed");
                
                
         orders.perform_cmd(async {
                RequestState::send(request)
                    .await
                    .map_err(|e| e.message().to_string())
                    .map_or_else(Msg::ValidateEmailFailed, Msg::ValidateEmailSucceed)
            });          

```

#### RequestState is a wrapper for making request and getting error message from backend 

```rust 
impl<T: 'static + DeserializeOwned> RequestState<T> {
    /// Send the request with `fetch(request).await?;` and catch every possible
    /// `FetchError` to ClientError;
    #[allow(clippy::needless_lifetimes)]
    pub async fn send<'a>(request: Request<'a>) -> Result<T, ClientError> {
        let response = fetch(request).await?;
        let status = response.status();

        if status.is_error() {
            let text = response.text().await?;
            return Err(ClientError::new(&text, &status.code.to_string()));
        }
        Ok(response.json().await?)
    }

    pub fn is_loading(&self) -> bool {
        match self {
            RequestState::IsPending(is_loading) => *is_loading,
            _ => false,
        }
    }
}
```

#### The data Model 

```rust

use serde::{Deserialize, Serialize};
#[derive(Deserialize, Serialize)]
pub struct FormEmail {
    email: String,
}
```

I do not know what is happening there and why my object is not passed like before as json. 

Any clue or indication on what I could do to fix it ?

fosskers:
~~I would love it if there were a better way to canonically turn a `JsValue` into a `String`.~~ There is!

Closes #619

### DIFF
--- a/src/browser/web_socket.rs
+++ b/src/browser/web_socket.rs
@@ -132,10 +132,9 @@ impl WebSocket {
     ///
     /// Returns error when JSON serialization or sending fails.
     pub fn send_json<T: Serialize + ?Sized>(&self, data: &T) -> Result<()> {
-        let data = JSON::stringify(&swb::to_value(data)?)
+        let data: String = JSON::stringify(&swb::to_value(data)?)
             .map_err(|_| WebSocketError::ConversionError)?
-            .as_string()
-            .ok_or(WebSocketError::ConversionError)?;
+            .into();
         self.send_text(data)
     }
 
--- a/src/browser/web_socket.rs
+++ b/src/browser/web_socket.rs
@@ -2,6 +2,7 @@
 
 use crate::app::Orders;
 use gloo_file::FileReadError;
+use js_sys::JSON;
 use serde::Serialize;
 use serde_wasm_bindgen as swb;
 use wasm_bindgen::{JsCast, JsValue};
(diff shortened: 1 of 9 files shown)


src/browser/web_socket.rs
```diff
@@ -131,7 +132,8 @@ impl WebSocket {
     ///
     /// Returns error when JSON serialization or sending fails.
     pub fn send_json<T: Serialize + ?Sized>(&self, data: &T) -> Result<()> {
-        let data = swb::to_value(data)?
+        let data = JSON::stringify(&swb::to_value(data)?)
+            .map_err(|_| WebSocketError::ConversionError)?
             .as_string()
             .ok_or(WebSocketError::ConversionError)?;
```
RReverser, line 138: Since `JSON::stringify` returns a `JsString`, you can instead use infallible `String::from(...)` on it (or `.into()` equivalent).

fosskers: That's way better, thank you. I must have somehow missed that those trait instances existed.

fosskers:
```rust
impl<'a> From<&'a JsString> for String {
    fn from(s: &'a JsString) -> Self {
        s.obj.as_string().unwrap_throw()
    }
}
```
I thought `From` was never supposed to panic.

RReverser: It won't because type is statically known here.

RReverser: Also, `.unwrap_throw()` doesn't panic in Rust terms, it's much more lightweight method as it immediately throws JS exception that unwinds entire JS stack.

PR #622, opened 2021-09-17
Fix Issue#232 link in Readme-Why not Seed section

badhrink:
It was redirecting to:
https://github.com/seed-rs/seed#232

but I think what was meant to be is this:
https://github.com/seed-rs/seed/issues/232


fosskers (approved):

fosskers:
CI failing due to a lint unrelated to this PR. I'll fix that first and then run your CI again.

PR #624, opened 2021-09-19
Allow passing owned values to Request.body (again)

glennsl:
Seems it's sometimes convenient to be able to pass owned values as well, so this reverts the earlier change from #606 and adds `body_red` instead. It would be nice if it was possible to have `body` take an `impl Into<Cow...>`, but unfortunately it doesn't seem to be. At least not without some convoluted proxy type implementing `From` specifically for `JsValue` and `&JsValue` as anything meoe generic would have the `impl`s conflict.

Also, I guess this is an unbreaking change, in a sense.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,12 +24,12 @@
 - Fixed an issue in vdom where inputs with invalid contents being cleared on Firefox.
 - Added `fetch::form_data::FormData` and `Request.form_data`.
 - Added `sl_input` to the `custom_elements` example.
-- [BREAKING] Changed `Request.body` to take its argument by reference.
 - Adapted to Rust 1.53.0.
 - Removed internal `serde_json` usage in favour of `serde-wasm-bindgen`. This reduces final binary size for downstream users.
 - [BREAKING] Removed the deprecated `browser::service::fetch` module.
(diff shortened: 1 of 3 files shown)


fosskers:
> Seems it's sometimes convenient to be able to pass owned values as well

For instance? I'm interested in the motivation for this.

glennsl:
In our case it's quite a complex scenario where we use our own request data structure because we need it for offline support, to be able to serialize and store it in IndexedDB and then sync it once back online. In one case we can store the body directly as a `JsValue`, and could just borrow that (unless we'd like our request data to be consumed in the process), but in other cases we can't, so we have to unify the two at some point which leads to some awkwardness because something has to own the body until it's been sent. It would just be a bit more convenient if that something was seed's `Request` so we only had to pass that around.

fosskers:
Thanks for the clarification. Let's see what @MartinKavik thinks. I know that he's busy at the moment though, so for the time being are you okay with pinning your personal `seed` to this PR?

glennsl:
Yeah, no rush. We have it working with the current master as well, so this is just to make the API more convenient. And non-breaking. Thanks for looking it over!

flosse:
@glennsl What is the state of this PR? Do you think we can merge it before we release v0.9.0?

PR #626, opened 2021-09-28
feat(virtual-dom): blanket impl of IntoNodes for Option

glennsl:
This replaces the specific `impl`s of `IntoNodes` for `Option<Node<Ms>>` and `Option<Vec<Node<Ms>>` with a blanket implementation for `Option<T> where T: IntoNodes`. This isn't just for convenience, as it also solves the problem of not being able to implement `IntoNodes<Ms>` for `Option<T>` if `T` does not include `Ms` as a type parameter, since that violates the orphan rule (`E0210``).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@
 - [BREAKING] Removed the deprecated `browser::service::fetch` module.
 - Element macros like `div!` can now contain `Iterator`s inside of `Option` values. Previously only one or the other was possible.
 - Add method to return detailed error response from server with `FetchError`.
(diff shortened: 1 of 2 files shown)


glennsl:
This is actually a breaking change, as it might conflict with local `impl`s. Updated the change log to reflect that.

flosse:
LGTM!

fosskers (approved):

MartinKavik:
The group of Seed masters doesn't need to wait for me to click the Rebase button next time :) (At least @fosskers should have permissions to merge PRs.)

fosskers:
@MartinKavik Yup, just wanted to let it sit an extra day just in case. Thanks for merging.

Gam3356 (commented):
Thank you guys! Willie Parish IPHONE 12 Pro max

PR #627, opened 2021-09-29
feat(virtual_dom): add `on_insert` event triggered on DOM insert

glennsl:
This is an attempt at adding an `on_insert` "attribute" to `El`. The purpose of this is to simplify the initialization of inserted dom elements, as the current approach of using `el_ref` and `orders.after_render` can get very hairy if the element is inserted (and re-inserted) based on some non-trivial condition. Using `on_insert` to handle the lifecycle for you ensures the element is always initialized. It also removes a bit of boilerplate even in the trivial scenarios, which is nice.

My understanding of the virtual dom is a bit shaky though, hence the draft. I'm pretty sure I'm overdoing the triggering, as I've just added triggering wherever `ElRef`s are being set.

I've also not gotten the example to work correctly, as there seems to be some caching of nodes that prevent it from being re-inserted when toggled. Or I've missed adding a trigger to an insertion point somewhere. I will look further into it later, but would appreciate any advice you might have.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 - Element macros like `div!` can now contain `Iterator`s inside of `Option` values. Previously only one or the other was possible.
 - Add method to return detailed error response from server with `FetchError`.
 - [BREAKING] Added blanket `impl<Ms, T: IntoNodes<Ms>> IntoNode<Ms> for Option<T>`. This might conflict with local `impl`s of `IntoNodes`, but should make those unnecessary and safe to remove.
+- Added `on_insert` event on elements, triggered when they are inserted into the DOM.
 
 ## v0.8.0
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1578,6 +1578,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "on_insert"
+version = "0.1.0"
+dependencies = [
+ "seed",
+]
+
+[[package]]
 name = "once_cell"
 version = "1.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -137,6 +137,7 @@ members = [
     "examples/markdown",
     "examples/fetch",
     "examples/no_change",
+    "examples/on_insert",
     "examples/pages",
     "examples/pages_hash_routing",
     "examples/pages_keep_state",
--- a/examples/on_insert/Cargo.toml
+++ b/examples/on_insert/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "on_insert"
+version = "0.1.0"
+authors = ["Glenn Slotte <glenn@slotte.net>"]
+edition = "2018"
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+seed = {path = "../../"}
--- a/examples/on_insert/Makefile.toml
+++ b/examples/on_insert/Makefile.toml
@@ -1,0 +1,27 @@
+extend = "../../Makefile.toml"
+
+# ---- BUILD ----
+
+[tasks.build]
+alias = "default_build"
+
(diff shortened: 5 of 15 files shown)


glennsl:
As a follow-up to this I'm also considering an `on_size` event that would use `on_insert` to get the initial size, then set up a `ResizeObserver` to emit the same `Msg` on resize, eliminating the whole mess of dealing with that.

MartinKavik:
> My understanding of the virtual dom is a bit shaky though, hence the draft. I'm pretty sure I'm overdoing the triggering, as I've just added triggering wherever ElRefs are being set.

VDOM has been written before I started to maintain Seed. I've fixed some bugs and added few things. One or more contributors rewritten the patching algorithm.
If I remember correctly, I set ElRef in places discovered basically by trial & error because I had also problems to orient myself in the VDOM code. So don't rely on it too much.

> As a follow-up to this I'm also considering an on_size event that would use on_insert to get the initial size, then set up a ResizeObserver to emit the same Msg on resize, eliminating the whole mess of dealing with that.

I saw you've probably copied the ResizeObserver implementation I've written in JS from an example in the repo. I've recently written a Rust wrapper in MoonZoon - https://github.com/MoonZoon/MoonZoon/blob/983ea0ec87d5428df1ae096dd3eeb9273fb439c2/crates/zoon/src/resize_observer.rs - feel free to use it as you wish. Note - I've read somewhere it's better for performance to use one global ResizeObserver instead of one per element (however I still use multiple ResizeObservers in MoonZoon apps and I don't observer any problems.)

glennsl:
Thanks @MartinKavik. Yeah I just copied the entire `resize_observer` example and adapted that to have something to test with. I'll make something simpler for the proper example I think. For the `on_size` handler I figured I'd make proper `ResizeObserver` bindings, and find it very convenient that you've made some already! If you don't mind I'd like to copy those over to seed. Let me know if you want to make that commit yourself, or want some other kind of attribution for it.

glennsl:
I found that I can reduce the "wiring" to two places, that will cover all the current `ref setting`, and so I've refactored the code to also reduce the ref setting and event manager attachment to those two places, as well as cleaning up and factoring out common code in a few places in https://github.com/seed-rs/seed/pull/627/commits/220f3600261b861c444232b46c1e5be28bcbf4b5. If this seems too risky I can revert this commit and we can try out just `on_insert` for a while to make sure it works, then do this refactoring later.

Pre-refactor, the two locations that will cover everything without duplication are `virtual_dom_bridge::attach_el_and_children` (which was already setting refs) and `virtual_dom_bridge::insert_node`. `attach_el_and_children` is called from `patch::append_el` and `patch::replace_by_el`, as well as from `app::bootstrap_vdom`. `insert_node` is called only from `patch::insert_el`. These patch functions are in turn called by `patch::patch_els`, which was another function that set refs.

Because the wiring needs both the `El`, a `web_sys::Element` (rather than `web_sys::Node`) and a `Mailbox`, and `insert_node` has neither, I decided to make a wrapper function `insert_el_and_children` as an analog to `attach_el_and_children`, and moved some of the logic from `patch::insert_el` in there. It seems to me that `virtual_dom_bridge` is a bit fat and fuzzy on the edges already, and this doesn't improve that. but it does make it more consistent I think.

I also made `attach_children` a bit more general, so it could be more widely reused and removed some duplicate code, and added the `wire_up_el` function for the common wiring code.

glennsl:
I still need to make a proper example, add documentation and ought to do some more testing, so this is still a draft. But feel free to do a review of the `virtual_dom` code, as that should be done.

glennsl:
Made a few improvements to `on_insert` and simplified the example. Should be good for review now!

glennsl:
Forgot the change log. Again...

flosse:
@glennsl What is the state of this PR? Do you think we can merge it before we release v0.9.0?

flosse:
closing this in favor of #665

PR #657, opened 2022-03-22
Fixes

flosse:


flosse:
@fosskers or @MartinKavik would you mind to look at this or give me some access to that I can continue?

src/shortcuts.rs
```diff
@@ -74,12 +87,13 @@ macro_rules! with_dollar_sign {
 ///     }
 /// }
 ///
-/// fn view(model: &Model) -> Node<Msg> {
+/// fn view(mdl: &Model) -> Node<Msg> {
 ///     a![
-///         attrs!{ At::Href => Urls::new(base_url).home() }
+///         attrs!{ At::Href => Urls::new(mdl.base_url.clone()).home() }
```
MartinKavik, line 92: Perhaps `&mdl.base` would work here instead of cloning (haven't tried it, mark as resolved if not possible/not important)

MartinKavik:
> would you mind to look at this or give me some access to that I can continue?

@flosse Access given, write me a DM on Discord if it doesn't work or when there are any other similar problems.

flosse:
> @flosse Access given, write me a DM on Discord if it doesn't work or when there are any other similar problems.

:pray:

PR #659, opened 2022-04-03
Fix 619

flosse:
This is a quick fix but I'd prefer to use a feature flags like `swb-json` that is disabled and use `serde-json` instead by default (see #660).


fosskers:
Thank you for this.

PR #674, opened 2022-04-14
Fix: make fetch::JsonError public

### LINKED ISSUE #673: v0.9.1: browser::json::Error is inaccessible
As a consequence `FetchError::JsonError(_)` cannot be handled appropriately.

flosse:
should fix #673


Gam3356 (commented):
Thank you guys! Willie Parish IPHONE 12 Pro max

PR #691, opened 2022-06-18
Rust 1.61

fosskers:
I couldn't resolve the branches in #689 after editing it locally, so I've opened this separately. 

Closes #689
Closes #688


flosse:
Cool, thanks a lot!

PR #693, opened 2022-06-18
Fix ws json decoding

flosse:


wuerges:
Confirmed it solved my issue here.
Thanks!

Issue #701, opened 2022-07-27
[0.8.x] WebSocket and too early dropped closure


wkordalski:
Commit 08c4ce15255ac8d03df5315796e7b63e624e2987 introduced some bug:

```
Uncaught Error: closure invoked recursively or destroyed already
    at imports.wbg.__wbindgen_throw
    at wasm_bindgen::throw_str::ha13545fa78d0389b
    at <dyn core::ops::function::Fn<(A,)>+Output = R as wasm_bindgen::closure::WasmClosure>::describe::invoke::he801481bf9a481f4
    at __wbg_adapter_42
    at WebSocket.real
```

I suppose, one of the unused fields were used to prevent from dropping the closure mentioned above.
Rust says the field is unused, because its only use is to run `drop()` in the right moment.
We should mark that field with `#[allow(unused)]` (or some more specific lint from the "unused" group) instead of removing it.

flosse:
This should be fixed in `v0.10.0` (see websocket example). If it is still not working, feel free to re-open this issue.

PR #702, opened 2022-07-28
Revert "Remove unused fields"

wkordalski:
This reverts commit 08c4ce15255ac8d03df5315796e7b63e624e2987.

The fields are considered as unused by `rustc`, however they might be stored to be dropped in the right moment.

In the future we should analyze which field we can really delete, however I think we can do without such clean-up as Seed 0.8 is not the current version of Seed.

See: https://github.com/seed-rs/seed/issues/701

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -4,7 +4,7 @@ use crate::browser::{
     util::{self, window},
     Url, DUMMY_BASE_URL,
 };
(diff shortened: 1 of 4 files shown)


flosse:
Should be obsolete since websocket connections are handled outside since v0.10.0.
Feel free to re-open this PR if there are still problems.

